PR #9, opened 2015-02-16
Adding custom validators support

lionel-panhaleux:
Hi,

My name is Lionel and I'm working at a french firm called Polyconseil. I found your yamale library the other day, and it is really well made. Yamale could be very useful for our new YAML configuration architecture. I took the liberty to make this pull request to add a feature we need : the ability to add some custom validators to the default yamale validators. I used a pretty simple "date" validator for testing purpose, but we have other custom format we would like to validate, which can not be added to the default validators because they are not so standard. I hope I'll have the opportunity to make some other PRs in the near future to help and improve this nice tool.

Kind Regards

PR #10, opened 2015-02-18
Allow custom validators

blopker:


blopker:
Adding to https://github.com/23andMe/Yamale/pull/9

PR #12, opened 2015-03-03
Date and Datetime validators + universal constraints

lionel-panhaleux:
Hi,

I wanted to add date and datetime validators, and I figured since they are standard libraries and timestamps and dates are already directly handled by pyyaml, they had their place in the library.

Once I got those validators, I figured it would be nice to be able to specify min and max constraints on them and I tried to pull something up to render some constraints "universal", meaning they could adapt to any compatible type (e.g. any comparable type could, even a custom one, can get a min/max constraint). I think it matches the zen of python, but of course the implementation is debatable.

What do you think ?


blopker:
Hey Lionel, good idea. I'll check this out tomorrow!

blopker:
Hey! I was finally able to take a look at this, sorry for the wait.

I added the day and timestamp validators, but I had to remove the 'default' stuff. As of right now the default feature is out of scope for Yamale. It's only meant to be a validator and not a deserializer, which means users aren't really supposed to be using the Data objects after they are validated. Although I see the appeal of this functionality. 

I think when/if Yamale 2 comes around it will target this use case.

Cheers! :beers:

PR #14, opened 2015-03-15
fix an issue where the custom validators were not propagating to the includes. Add tests for the case.

goodwillcoding:


blopker:
Sweet! Thank you.

PR #20, opened 2015-09-27
Fix included lists

sebdiem:
I added a test which fails: when multi-level lists are included, elements cannot be retrieved using the keys "0.0", "0.1" etc…
I propose a fix to handle these kind of cases, which can be part of a valid schema.


blopker:
Merged in #21, thank you!

PR #23, opened 2015-12-16
Path Validation, Yo

leycec:
This pull request adds support for validating YAML objects as filesystem paths, complete with unit tests, Python docstrings, and Markdown documentation.

## That's Nice, Dear.

Specifically, four new `Validator` classes have been added.

### Pathname: Thy Name Is Hardship

The new `Pathname` validator subclasses the existing `String` validator. As the name suggests, it validates passed objects to be syntactically valid absolute or relative pathnames for the root filesystem of the current OS _regardless_ of whether those paths actually exist.

Simple, right? I thought so. In fact, pathname validation turns out to be:

* **Conceptual non-trivially,** in that no one including [some guy named Guido](https://en.wikipedia.org/wiki/Guido_van_Rossum) appears to know how to do it cross-portably. Or at all. Actually.
* **Technically feasible,** in that this pull request does it cross-portably in exactly 20 lines excluding whitespace and assertions. (_You know. Just sayin'._)

See [this Stackoverflow answer](https://stackoverflow.com/a/34102855/2809027) for a detailed breakdown. Here's the twenty-second synopsis:

1. If the passed object is either _not_ a string or is the empty string, this is an invalid pathname.
1. Else, split the passed object on the path separator into path components (e.g., `/Autumn/Aurora` into `['', 'Autumn', 'Aurora']`) _after_ ignoring the drive specifier prefixing this pathname (e.g., the `C:` in `C:\Forgotten\Legends`) if any.
1. For each such path component:
   1. Concatenate the absolute path of a directory known to exist (e.g., the root directory `/` under POSIX-compatible OSes) with this path component into a new pathname.
   1. Attempt to stat this pathname.
   1. If doing so raises any of the following exceptions, this path component and hence the passed object is invalid:
      * Under Windows, instances of `WindowsError` whose `winerror` attribute is `123` (i.e., `ERROR_INVALID_NAME`).
      * Under all other OSes:
        * For pathnames containing null bytes (i.e., `'\x00'`), instances of `TypeError`.
        * For pathnames containing path components longer than 255 bytes, instances of `OSError` whose `errcode` attribute is:
          * Under SunOS and the *BSD family of OSes, `errno.ERANGE`. This appears to be an OS-level bug, otherwise referred to as "selective interpretation" of the POSIX standard.
          * Under all other OSes, `errno.ENAMETOOLONG`.

It's worse than it sounds. _No, really!_

Unit tests exhaustively exercising this functionality have been added _and_ successfully run under both Python 2.7 and Python 3.4 for both Windows and Gentoo Linux. Emphasis on "exhausting."

Front-facing `README.md` documentation _and_ inline [`numpydoc`](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt)-style docstrings documenting this validator have also been added. If Yamale uses a different docstring style (e.g., [`googledoc`](http://google.github.io/styleguide/pyguide.html#Comments)), I am apologetic but ultimately unrepentant.

### Path, File, and Directory (Because We Can)

The new `Path` validator subclasses the above `Pathname` validator. It validates passed objects to be the absolute or relative pathnames of existing paths of any type (e.g., files, directories, device nodes, symbolic links, _whatever_).

The new `File` and `Directory` validators subclass the above `Path` validator. They validate passed objects to be the absolute or relative pathnames of existing files or directories respectively. No one saw that coming.

Unit tests and documentation have also been added for these validators.

Thus was paradise lost.

Paradise Lost

## An Orgasmic Smorgasbord of Utility Functions

The existing `yamale.schema.util` module defines a number of general-purpose utility functions of interest elsewhere (e.g., `isstr()` in the `yamale.validators.validators` module). So, it's been moved to `yamale.util`. Two new utility functions required by the above validators have also been added:

* `is_os_windows_vanilla()`, testing whether the current OS is **vanilla Microsoft Windows** (i.e., _not_ running the Cygwin POSIX-compatibility layer). We don't need to test this terribly often – but when we do, it's invaluable.
* `get_subclasses()`, a generator recursively yielding all transitive subclasses of the passed class (in depth-first order). _Yeah._ It's `type.__subclasses()__` on topical steroids. Whereas `type.__subclasses()__` yields only the direct subclasses of a class, `get_subclasses()` yields all indirect subclasses as well. This is critical, as the above validators are _not_ direct subclasses of the `Validator` class and hence cannot be added to the `DefaultValidators` dictionary via the old `Validator.__subclasses()__` approach.

Too much information, maybe? But the more you know... the less you seem to know.

## Appveyor – A Sane Testbed for an Insane Platform

**NOTE: Incoherent, unrelated gibberish follows.** Yet it must be said.

`<rant>`

I deride, fear, and loathe (in equal measure) all things touched by the billowing tentacles and monolithic bile of the Great Slime Mold Microsoft. Against my better nature, I used Windows to test this pull request. _It felt wrong._

Microsoft

I'd like to contribute to Yamale again, unless I have to interface with Windows again to do it. I can't be the only one. (Anybody? _Anybody!?_) Therefore, I have a malingering request:

> Please register Yamale with the Windows-specific continuous integration (CI) service [Appveyor](http://appveyor.com).

It's the `tox` of the sordid Windows world. It's also free-as-in-beer for open-source projects.

We use Appveyor for automated Windows testing of [PyInstaller](https://github.com/pyinstaller/pyinstaller), a heavy-weight cross-platform open-source project with both a pure-Python component (PyInstaller proper) _and_ a C component (PyInstaller's bootloader) _and_ a medley of platform-specific dependencies. Despite this wackiness, Appveyor just works. I haven't had to filthify my soul with Windows even once.

**Appveyor:** the undiscovered Eighth Wonder of the World.

`</rant>`

## To Be Subsequently Done By Someone Else Who Is Not Me

In a word, **constraints.**

This pull request does _not_ add fine-grained support for validating filesystem constraints (e.g., POSIX-style path readability, writability, or executability for user, group, or other). It shouldn't be that unspeakably hard for someone else (_...who is not me_) to do that at some future date (_...which is not now_).

Two hard-boiled approaches come to mind:

1. Add additional `Validator` subclasses inheriting validators added by this pull request. For example, a new `Symlink` class validating passed objects to be existing symbolic links (dangling or not) by subclassing the new `File` validator.
1. Add additional `Constraint` subclasses specific to validators added by this pull request. For example, a new `Readable` subclass with keyword `readable` applicable only to the `Path` validator and subclasses thereof (e.g., `File`, `Directory`) validating passed paths to be... readable! (*surprise*)

## Is Yamale Dead?

Because, you know, it kind of looks that way. The bloated blood-flies are gathering to feast; the slathering army of corpse maggots assembles nearby.

The lack of commits in nearly two months is concerning. I _was_ contemplating using Yamale in production code. Despite submitting this pull request, I'm beginning to rethink that choice. Or... maybe not?

Probably not. [PyKwalify](https://github.com/Grokzen/pykwalify) – Yamale's _much_ fatter, uglier, and less Pythonic step-brother – is the only viable alternative, and it's less frequently updated than even Yamale! (Last commit: [September 30th](https://github.com/Grokzen/pykwalify/commit/5730eac404f8f7888b92321ded9be278744870c8)). It's kind-of sad.

I can only assume that either **(A)** nobody cares about YAML validation under Python (_inconceivable! I care and you had better too_) or **(B)** YAML validation under any language is simplistic and stable enough to not warrant 2GB updates every other day (_...possible, however unlikely_).

## The Future Beckons with a Sick Smile

Thanks for all the fishy YAML validation. Regardless of whether Yamale is dead or merely on Github life support, you've done a bang-up job.

Thanks! :smile:


blopker:
Hey @leycec, love the tests and we can always use more validators. This is a lot to take in though so I'll have to check it out tomorrow.

On Yamale being 'dead', it's not. We use it in production to validate thousands (maybe tens of thousands) of yaml files hundreds of times a day, including files which are classified as medical devices. The reason for the lack of commits is simply that it was created with a well defined interface which it implements without issue. We just haven't needed anything more from it.

As far as Windows support goes (yes, you are the only one :smile:), I'll look in to it. However, it does increase maintenance burden on us. I'll have to evaluate how much.

Cheers! :beers:

leycec:
> This is a lot to take in though so I'll have to check it out tomorrow.

That... was a _really_ fast reply. I'm tired as a decaying carcass, so I may not get back to you for a few days.

> As far as Windows support goes (yes, you are the only one :smile: )...

I knew it.

> However, it does increase maintenance burden on us.

Microsoft has a way of doing that. You're in luck, though: Yamale tests already pass under Windows! (Was that intended? Not that I'm complaining.)

**Also, [Appveyor](http://appveyor.com).** It's bad-assery. You just register Yamale as a public Appveyor project, define a top-level [`appveyor.yml`](https://github.com/ogrisel/python-appveyor-demo/blob/master/appveyor.yml) file, and away you go. Github integration handles the rest (e.g., ensuring that Appveyor tests are run on all prospective pull requests).

yamale/util.py
```diff
@@ -86,3 +95,39 @@ def get_iter(iterable):
         return iterable.items()
     else:
         return enumerate(iterable)
+
+
+def get_subclasses(cls, _subclasses_yielded=set()):
```
blopker: Mutable default arguments in Python: http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

yamale/validators/tests/test_validate.py
```diff
@@ -22,6 +38,135 @@ def test_string():
...
 
+def test_validator_pathname():
+    '''
+    Unit test the `Pathname` validator.
+    '''
+
+    v = val.Pathname()
+
+    # Directory guaranteed to exist. See Pathname._is_valid().
+    root_dirname = os.environ.get('HOMEDRIVE', 'C:') \
```
blopker: Line continuations are fragile, best to use an `if` block instead.

yamale/validators/validators.py
```diff
@@ -21,7 +12,236 @@ class String(Validator):
     constraints = [con.LengthMin, con.LengthMax, con.CharacterExclude]
 
     def _is_valid(self, value):
-        return isstr(value)
+        return util.isstr(value)
+
+
+class Pathname(String):
```
blopker: It seems like inheriting from `String` doesn't give `Pathname` much here. It's easy enough just tor reuse `isstr`. Inheriting also means you bring `String`s `constraints` along which doesn't really apply to `Pathname`.

yamale/util.py
leycec: Classic newb mistake. :sheep:

yamale/validators/tests/test_validate.py
leycec: > Line continuations are fragile...

Readability concerns I grok, but... _fragile_? I've never had a line continuation fall down on me, but have encountered this opinion in the Python community before. There's even a [rejected PEP](https://www.python.org/dev/peps/pep-3125) arguing for the removal of explicit line continuations from the language altogether. I'm not quite sure what drives this, but suspect that Python developers attempting to distance themselves from lower-level languages (e.g., C, `sh`) and competitors (e.g., Ruby) accounts for a bit of it.

_Anyway._ I'll happily do it, of course; no troublesome complaints from me. Merely curious! :octocat:

yamale/validators/validators.py
leycec: > Inheriting also means you bring `String`'s constraints along which doesn't really apply to `Pathname`.

That's _totally a valid concern. I'll reinherit `Pathname` directly from `Validator` then. Thanks for the thoughtful heads up!

yamale/validators/tests/test_validate.py
blopker: It stems around the fact that an invisible character can break the code. Now, at first this seems silly since Python is a whitespace significant language, but this whitespace happens to be at the end. Maybe fragile is a strong word, but I know I'd be confused if I accidentally added a space and code started breaking with no visible change.

yamale/validators/validators.py
blopker: Don't make any changes yet, I still have to talk to people about Windows support. It's not something we normally do.

leycec:
**It is done.**

This doesn't really pertain to this pull request, but the generalized definition of the `yamale.validators.validators.DefaultDictionary` dictionary implies that the current approach for defining custom validators _could_ be simplified quite a bit. Currently, users have to slog through a bit of Java-ish boilerplate code to register custom validators: e.g.,

```
from yamale.validators import DefaultValidators, Validator
validators = DefaultValidators.copy()  # This is a dictionary
validators[Date.tag] = Date
schema = yamale.make_schema('./schema.yaml', validators=validators)
```

Since `DefaultValidators` now implicitly detects _all_ `Validator` subclasses, the above could be reduced to a single line of code: e.g.,

```
schema = yamale.make_schema('./schema.yaml')
```

"What miracles you speak of?", you are probably thinking. Some miracles are simpler than others, however. Just delay the definition of `DefaultDictionary` to the first `yamale.make_schema()` call (rather than immediately defining that dictionary in the `yamale.validators.validators` module) and the black magic of `util.get_subclasses()` will do the rest.

blopker:
I agree, adding custom validators is a little clunky. It was in response to another pull request and not part of the initial development. Oh well. However, it would be surprising to me as a user if simply subclassing `Validator` added my class to `DefaultValidators` as I (as a user) wouldn't consider it a default. Also that kind of magic seems more 'Ruby-ish' to me and not explicit enough for Python.

Maybe something like would be better than what we have:
```python
schema = yamale.make_schema('./schema.yaml')
schema.addValidator(tag, validator)
```

As for the bulk of this pull request, I heard back from some of the other engineers here. Unfortunately we don't have the bandwidth to maintain Windows specific code (who does?). Without the Windows-specific stuff the new `Validators` are not much more than the `String` validator with some constraints about null characters and length. So since it's easy to add your own Validators we won't be able to include these in core Yamale. However you've made some other improvements which I have checked out and am in the process of merging, like the improved subclass finder.

That being said, I'm interested in Appveyor as there's nothing in intrinsically in Yamale that's OS specific. I'll see about adding that to our CI.

Cheers! :beers:

leycec:
> Unfortunately we don't have the bandwidth to maintain Windows specific code (who does?).

The code is unit tested to work, so no bandwidth on your part is required.

Wouldn't simply ignoring Windows-specific logic suffice? It wouldn't be any skin off your teeth. After all, Python _is_ an interpreted language. Conditional branches that are never run had might as well not exist. If a Windows-specific branch were (hypothetically speaking) to fail in the indeterminate future, well... _who cares?_ That wouldn't impinge on the usability of Yamale under non-Windows OSes.

> Without the Windows-specific stuff the new `Validators` are not much more than the `String` validator with some constraints about null characters and length.

_Yeah_. Extreme over-trivialization much? Pathname validation is infeasible in the general case unless you perform a kernel call. Which this pull request did. Matching null bytes and path components longer than 255 bytes only works if you confine yourself to the subset of `ext2`-like filesystems, even ignoring Windows. (Which still accounts for [89.14%](https://en.wikipedia.org/wiki/Usage_share_of_operating_systems) of the desktop market. You know. It's _kind-of_ important.)

A few obvious examples where brute-force string handling falls down and coughs up a lung:

* [Files-11](https://en.wikipedia.org/wiki/Files-11) (i.e., ODS-5), the filesystem for [OpenVMS](https://en.wikipedia.org/wiki/OpenVMS). **Super-important in the super-computing and cluster space.** Files-11 imposes a number of interesting constraints on pathnames, including:
  * Path components no longer than **235 bytes.** _Not_ 255.
  * Pathnames at most **4Kb.**
* [Reiser4](https://en.wikipedia.org/wiki/Reiser4#History_of_Reiser4) – which, interestingly, is still being actively maintained. Reiser4 is quite a bit more lenient than most `ext2`-like filesystems, permitting:
  * Path components no longer than **4Kb**, plus or minus a few bytes. Again, _not_ 255.
* [Novel Storage Services (NSS)](https://en.wikipedia.org/wiki/Novell_Storage_Services), the filesystem for the [Novel NetWare](https://en.wikipedia.org/wiki/Novell_NetWare) OS. Once super-important in the service space, but admittedly long in the tooth today. Still, the filesystem has a number of novel and bizarre features. Most notably:
  * **Mimicry.** NSS can actually masquerade as a number of different filesystems (which it refers to as "namespaces"), including FAT32, NTFS (the default), ext2, HFS, and HFS+.
* [UDF](https://en.wikipedia.org/wiki/Universal_Disk_Format), the filesystem of choice for live Linux distributions on rewritable optical media everywhere. In this case, UDF requires:
  * Pathnames at most **1Kb.**

In the general case, naive string handling gets you nowhere.

> So since it's easy to add your own Validators we won't be able to include these in core Yamale.

That's... a disappointing decision, to say the least. I spent quite a bit of time on this pull request. This doesn't exactly encourage future participation on my part.

That's also an interesting rationalization for shutting down volunteer contributions. It doesn't particularly hold up to scrutiny. I mean, "It's easy for you to monkey-patch in your own changes. So just go do that! kthx." _Really?_ How does my monkey-patching an application-specific solution actually benefit anyone else? That was the whole point of this tiresome escapade, after all: **to benefit others.**

General-purpose pathname validation is non-trivial. What's the point in requiring _every_ cross-platform application validating YAML files containing pathnames to reimplement that from the ground up?

I mean, I sort of get it. Window's isn't a priority. Very well. It's not a priority for me or most open-source advocates either. I'm an ardent Gentoo Linux evangelist, you know? Open-source doesn't get more hardcore. But why dogmatically leap from the reasonable position of "Windows isn't a priority." to the extreme position of "Windows logic ist alles verboten!"?

**Anyway.** We're done here. Best of luck with the future of Yamale.

As our cross-portable application supporting Linux, OS X, and (_...wait for it_) Windows requires YAML validation, I may elect to publicly fork Yamale at some future date. Windows support is a non-optional hard requirement for us. You'll certainly be notified if and when this occurs.

PR #25, opened 2016-03-10
Add silent flag to validate() to allow or disallow printing includes

cdparks:
Currently, when `_validate_include()` is called, included validators are printed to `stdout`. This is noisy and slows down validation for deeply nested schemas.

For example, suppose I have this schema:

```yaml
id: int(min=1)
title: str()
content: list(include('headingT'), include('paragraphT'))
---
headingT:
  heading: str()

paragraphT:
  paragraph: str()
```

and this file:

```yaml
id: 1
title: This is a title
content:
- heading: This is a heading
- paragraph: This is some text
```

Validating this file will produce the following output:

```plaintext
>>> import yamale
>>> _ = yamale.validate(yamale.make_schema('schema.yaml'), yamale.make_data('data.yaml'))
('heading', String((), {}))
('paragraph', String((), {}))
('heading', String((), {}))
('paragraph', String((), {}))
>>>
```

This change turns this off by default:

```plaintext
>>> import yamale
>>> _ = yamale.validate(yamale.make_schema('schema.yaml'), yamale.make_data('data.yaml'))
>>>
```

Pass `silent=False` to `validate()` to get the old behavior:

It's possible the `print()` ([see this line](https://github.com/23andMe/Yamale/blob/master/yamale/schema/schema.py#L141)) was just a debugging artifact, in which case, it would probably be better to just delete it.

### DIFF
--- a/yamale/schema/schema.py
+++ b/yamale/schema/schema.py
@@ -36,18 +36,18 @@ class Schema(object):
                 raise SyntaxError(str(e) + ' at node \'%s\' in file %s' % (key, self.name))
         return schema_flat
 
-    def validate(self, data):
+    def validate(self, data, silent=True):
         errors = []
 
         for key, validator in self._schema.items():
-            errors += self._validate(validator, data, key=key, includes=self.includes)
+            errors += self._validate(validator, data, key=key, includes=self.includes, silent=silent)
 
         if errors:
             header = '\nError validating data %s with schema %s' % (data.name, self.name)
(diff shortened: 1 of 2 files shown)


blopker:
This print statement was an oversight. I've just removed it in #26

cdparks:
Thanks for the quick response!

blopker:
No problem, thanks for pointing this out. :+1:

PR #32, opened 2017-07-18
Look for schema in different folders.

Roussel-C:
"schema" file was always located in the "data" directory thus preventing the user to specify different folders.

The current proposal is to perform the following steps:
  - check if "schema" is a path to an existing file (eg: "./myschema.yaml" or "/home/user/schema.yaml")
  - Else, try to see if it's an expression (eg: "./schema*.yaml" or "/home/user/schema*.yaml") and take the first valid file (no matter its extension).
  - Else, look in "data" folder for all files that match schema expression and take the first valid file.
This also make it possible to use relative paths (eg: '../schema.yaml').

When everything fails then raise an ValueError (If we have no schema, there is no validation).

__BTW__: ".yml" becomes a valid extension.

### DIFF
--- a/yamale/command_line.py
+++ b/yamale/command_line.py
@@ -28,13 +28,28 @@ def _validate(schema_path, data_path):
 
 
 def _find_schema(data_path, schema_name):
-    if not data_path or data_path == '/' or data_path == '.':
-        return None
-    directory = os.path.dirname(data_path)
-    path = glob.glob(os.path.join(directory, schema_name))
-    if not path:
-        return _find_schema(directory, schema_name)
-    return path[0]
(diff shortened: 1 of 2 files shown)


blopker:
Hey @Roussel-C, thanks for the PR! I'll check it out.

This looks good, but in general you should you should create a proposal issue with the change you want to make first, but no code. That way if the maintainer has a different way to do it they can tell you before any code is written.

Cheers!

blopker:
OK, I made some modifications and put them here: https://github.com/23andMe/Yamale/pull/33

Can you check that this solves your issue?

Roussel-C:
Ok. Sorry for not asking. 
Yes! It solves my issue.

PR #44, opened 2018-11-19
feature/HEL-3992 Add kwarg for none

antfitch:
The problem:
If you have an optional node and you need it to have a value, there is no way to enforce this.

The solution:
Added a new kwarg called none. If none is true, a value is not required. If none is false, a value is required. By default, none is true for backwards compatibility.

A note:
I was initially going to add this as a constraint, but this was a problem because the code first checks if an optional node has no value. If no value is detected, constraints are not processed for the node. The solution would have required more code changes and just seemed a bit less elegant to me.

Take a close look at those tests! If you need more, let me know.

To test with product:
1. For product, load feature/HEL-3992
2. For yamale, load feature/HEL-3992
3. In yamale, follow these steps to create a local distribution of the yamale package and to load that package into product: 
https://23andme.atlassian.net/wiki/spaces/~111242565/pages/558170249/WIP+PIP+package+and+Vagrant
4. In product, go to data23/health/assessment/lark.
5. Add a new field to the schema. It could look like this:
mynode: str(required=False, none=False)
5. Add a new field to one of the other yamls in this directory that does not have a value. It could look like this:
mynode:
6. In product, run: (cd data23/health && make clean && make)
You should see an error pop up for mynode.


yamale/tests/test_meta_test.py
```diff
@@ -75,3 +75,21 @@ def runTest(self):
...
+        self.assertRaises(ValueError, self.validate)
+
+
+class TestGoodRequiredYaml(YamaleTestCase):
+    base_dir = data_folder
+    schema = 'meta_test_fixtures/schema_required_good.yaml'
+    yaml = 'meta_test_fixtures/data_required_good.yaml'
+
+    def runTest(self):
+        self.assertTrue(self.validate())
```
darreny23: Very minor, but it looks like you don't have a new line at the end of this file.  You should turn on auto new line for your editor.

README.md
```diff
@@ -188,6 +188,8 @@ Validators
 ----------
 Here are all the validators Yamale knows about. Every validator takes a `required` keyword telling Yamale whether or not that node must exist. By default every node is required. Example: `str(required=False)`
 
+You can require an optional node to have a value. A value is not needed by default. In this example, an optional node must have a value if the node is used: `str(required=False, none=False)`
+
```
blopker: Maybe:
```
You can also require that an optional value is not `None` by using the `none` keyword. By default Yamale will accept `None` as a valid value for a key that's not required. Reject `None` values with `none=False` in any validator. Example: `str(required=False, none=False)`.
```

yamale/tests/test_meta_test.py
antfitch: Thank you for the tip! Found the setting in VSCode and turned it on.

README.md
antfitch: @blopker, I've updated the README with the new text. I also updated the version from 1.7.1 to 1.7.2. If you want it to stay at 1.7.1, let me know.

blopker: Generally I'll handle the versioning, but this is fine. Thanks!

PR #46, opened 2019-02-09
Fix deprecation warnings

vladorf:
Hi, since python 3.3 ABCs for containers were moved to their own module and will be removed from original in 3.8. This PR fixes it, can you please look into it?


blopker:
Nice, thanks for this. I'll try and make a release this week.

vladorf:
Thanks for fast response, appreciate the effort you are putting into this

Issue #40, opened 2017-11-23
Any plan to add regular expression or pattern validation?


manishabardiya:
Hi,
Do you have any plan to add regular expression or pattern validation?

Thanks,
Manisha

blopker:
Hey @manishabardiya, I don't have much time for anything but bug fixes these days. However, if you'd like to make a PR I'll take a look.

manishabardiya:
Sure. I will connect you on that.

By the way, with current implementation of Yamale how do you handle such data yaml?
---
"demo1":
    layer:    "1" 
    dir: "/etc/abc"
...

where, demo1 is a token value, it can be anything so we don't have fixed key here. how do you handle such yaml with schema? Can you suggest?

Thanks,
Manisha

blopker:
Man, sorry I missed your question. There's currently no way to handle that. In my experience dynamic keys tend to be an anti-patterns. Many tools do not support it.

mildebrandt:
It seems the original issue has been fixed. If there is additional need or interest for validating dynamic keys, please open a separate issue and we can discuss further. Thanks!

PR #47, opened 2019-02-18
Regular expression validator

kangtastic:
Hi, here's my stab at adding a regular expression validator. I would've created a feature request but one exists already at #40.

The default fail message was modified, because `'foo' is not a regex` didn't make any sense, but further customization of the fail message to something like `'foo' is not a User ID` may or may not be something you want.

I considered allowing regex flags to be passed to `re.compile()` by using the validator like `regex(r'pattern', [r'pattern2',] flags=re.IGNORECASE, name='eggs')`, but that looked like it would've required adding the `re` module (or at least the various flags it defines) to the allowed (very) short list of `safe_globals` in `syntax.parser`. So... never mind that. :relieved:

I also updated the README to document the validator and fix a minor inaccuracy in the existing documentation for String.

Thanks for reviewing!

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,12 +208,13 @@ Validates strings against one or more regular expressions.
 - keywords:
     - `name`: A friendly description for the patterns.
     - `ignore_case`: Validates strings in a case-insensitive manner.
-    - `multiline`: `^` and `$` in a pattern match at the beginning and end of each line in a string in addition to matching at the beginning and end of the entire string.
+    - `multiline`: `^` and `$` in a pattern match at the beginning and end of each line in a string in addition to matching at the beginning and end of the entire string. (A pattern matches at [the beginning of a string](https://docs.python.org/3/library/re.html#re.match) even in multiline mode; see below for a workaround.)
     - `dotall`: `.` in a pattern matches newline characters in a validated string in addition to matching every character that *isn't* a newline.
 
 Examples:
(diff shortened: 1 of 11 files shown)


blopker:
This looks great! You are right, we want to keep the number of safe_globals very limited to keep the API minimal and not too Python specific. That being said, you could always add anther kwarg for each option you want to include (`ignore_case=True`).

The only thing this PR is missing: tests! Let me know if you have trouble getting the test suite running. I'd like to see both positive and negative cases.

kangtastic:
I've never written tests before, but I figured it out.

It turns out you can just pass in regex flags inline in the pattern string in Python 3, but I added three flags for Python 2 users.

Let me know if you have any more concerns or suggestions.

blopker:
This is looking really good. Thanks for the solid testing. I just had one suggestion. Also, you may want to rebase onto master.

kangtastic:
Yes, I may want to do that.

I think that about covers it on my end. Thanks for bearing with me, I'm still not very good at this :grinning:

blopker:
Sorry, I forgot to finalize my review. The one comment I had was to add some testing for the flags.

kangtastic:
Good idea. I wrote more test cases for the validator.

blopker:
Great work, thanks!

kangtastic:
Glad to contribute, and thanks for the feedback! I'll be integrating tests into my workflow in the future because of it.

Issue #45, opened 2019-01-31
IP and MAC type validation


ryanmerolle:
I think leveraging the netaddress module for IP and MAC address validation would be useful.  I get that you do not have time for anything but bugs, so I figured I post this feature request and possibly get to it when I have time myself.  Maybe even someone else might want to take a stab at this  as well.

blopker:
Hey @ryanmerolle, I like the idea, but yes, time is limited!

I'm happy for field a PR, but if you need this functionality, Yamale allows you to write your own validators without pushing upstream. I'll leave this open for now though. Thanks!

ryanmerolle:
Closed by #49

PR #49, opened 2019-03-08
Add IP and MAC Address Validator

supertylerc:
This adds a very basic IP address validator.  It uses `ipaddress` from the Python3 stdlib and the backported version for Python2.  I'm unsure if `ipaddress` is in a default install of Python2.7.x.  It was in mine, but it's also listed on PyPI as a backport, so I added it to dependencies anyway.

Anyway, this will let you validate IPv4 and IPv6 addresses, optionally enforcing the version.  You could extend this a _lot_, but I don't want to boil the ocean.

I know you can do custom validators, but this seems like something that might be common enough that it is worth having upstream.

I tried to make sure I did all the right things with testing.  Please let me know if I missed something.


supertylerc:
This would resolve part of #45, at least at a very basic level, although that issue specifically mentions `netaddr` whereas this PR uses `ipaddress` since it's in stdlib for Py3 and backported to Py2.7.

supertylerc:
I added functionality for validating MAC addreses as well, mostly piggybacking off of the Regex validator that already exists.  A library like `netaddr` would definitely do this better, but I was trying to keep from bringing in external dependencies given that there's probably only a subset of people who want these features, and if those people feel strong enough they could always write custom validators that used netaddr.

blopker:
Cool. Is there a reason you didn't want to use regex for the IP validation too? I looked and it seems available.

supertylerc:
@blopker I had a few!  :)

First and foremost, `ipaddress` is in the Python3 standard library and does this (and much more) already.  Although it isn't in the Python2 standard library, it is backported, and Python2 support ends 1 January 2020 anyway.

I would have also needed to write more code to do the same thing with regex.  IP addresses are also more difficult than MAC addresses when it comes to regex -- for example, each octet can be a number between 1 and 255, which isn't easy to express or read as a regex (to me).  Another option might be more code to check those things.

Finally, although I only supplied a `version` constraint in this PR, the `ipaddress` module provides a wealth of other interesting possibilities, such as confirming that an IP is link local or not, enforcing specific prefix lengths, determining if the address is private or public, multicast or unicast, is a teredo address, a 6to4 address, etc.  Lots of possible constraints, all of which I can see use cases for, and some of which I have use cases for myself (just haven't written the code to implement the constraint yet, though I hope to in the future).

supertylerc:
Ok.  I'm done writing more commits now.  :)  I'll update if you have feedback, but from my perspective, I'm "done."  I'll likely add more constraints to `ip()` in the future as I require them (just starting to build out some models that require this) and submit more PRs for that.

blopker:
Fair enough :)

I think it's fine to use `ipaddress`, but meet me half way: most people will not be using this validator. Making everyone install the `ipaddress` backport seems like just wasted resources (and maybe a security risk, not sure who maintains it), especially for us internally. We'd be installing an extra package hundreds of times a day for something we wont be using for a while if ever.

The alternative is to remove `ipaddress` from `setup.py` and do a conditional `import` in the validator itself catching `ImportError`. The catch would reraise the exception with a message about needing to install `ipaddress` for Python 2. A similar message would also be in the docs.

Whacha think?

yamale/validators/validators.py
```diff
@@ -166,6 +167,40 @@ def get_name(self):
...
+    tag = 'ip'
+    constraints = [con.IpVersion]
+
+    def _is_valid(self, value):
+        return self.ip_address(value)
+
+    def py2_ip_address(self, value):
+        try:
+            ipaddress.ip_interface(unicode(value))
+        except Exception:
```
blopker: In Python exceptions are quite slow and should only be used for exceptional cases, not normal control flow. 

I'd suggest adding an `to_unicode` function in `utils.py`, next to the `isstr`. For Python 3 it would just return `str(input)` for Python 2: `unicode(input)`

blopker: This comment isn't relevant to this exception BTW, but the one on line 188. Although there might be a better, non-exception way to do this one too.

supertylerc: I know it's a point of debate and contention, but a frequently cited rule for writing Pythonic code is EAFP, which typically seems to result in try/except instead of `if`s.  Assume the best (and default), and handle the cases where it isn't that.  I can definitely make the requested change, just explaining my logic for doing it this way.  I'll work on it this week.

supertylerc:
I can remove the dependency from `setup.py` and add it as a try/except.  However, I think I'd have to suppress the exception and not re-raise.  I haven't tested this specific case, but I would expect that you'll need that package no matter what since the re-raised exception would cause the application to exit.  Either I misunderstood what you meant or there's something else I don't understand (which is pretty common).

yamale/validators/validators.py
blopker: Yes, it is Pythonic to ask for forgiveness, however, that implies you did something wrong in the first place :)

In any case, you're still following that principle, just in a place that's less of a performance impact.

yamale/validators/validators.py
```diff
@@ -166,6 +167,40 @@ def get_name(self):
...
+            return False
+        return True
+
+    def ip_address(self, value):
+        try:
```
blopker: You can put the `import ipaddress` statement here, in the function. Python caches imports so re-importing has littte/no performance impact.

blopker:
The idea would be to only import the package when the user actually needs it. If it's not there there the program can't continue anyway, so it might as well crash with a helpful message.

yamale/validators/validators.py
supertylerc: Ah, yeah, I misunderstood previously.  Thought you meant at the top, but can definitely do in the function.  Thanks!

supertylerc:
@blopker I've updated to implement the requested changes.

yamale/validators/constraints.py
```diff
@@ -115,3 +118,22 @@ def _is_valid(self, value):
 
     def _fail(self, value):
         return self.fail % (value, self._failed_char)
+
+
+class IpVersion(Constraint):
+    keywords = {'version': int}
+    fail = 'IP version of %s is not %s'
+
+    def _is_valid(self, value):
```
supertylerc: I'm not particularly thrilled with this implementation of the `IpVersion` constraint.  It's almost the same as the `yamale.validators.validators.Ip.ip_address` method.  I implemented a version of it that deduplicated code by importing `yamale.validators.validators.Ip`, and that works fine for Python3 but not for Python2.  I guess Python3 handles recursive imports but Python2 doesn't?  Not sure.

I also thought about splitting the IP-specific bits into its own module, but I wasn't sure how you'd feel about that, so I just opted to leave things as-is and ask for some feedback.

yamale/validators/constraints.py
```diff
@@ -115,3 +118,22 @@ def _is_valid(self, value):
...
+        try:
+            import ipaddress
+        except ImportError:
+            raise ImportError("You must install the ipaddress backport in Py2")
+        try:
+            ip = ipaddress.ip_interface(to_unicode(value))
+        except ValueError:
+            ip = {'version': None}
```
blopker: Should this just return False? When the code gets to the next line it should blow up since `ip.version` wont work on a dictionary. Is there a test for this case?

yamale/validators/constraints.py
blopker: You could move both try/excepts to the utils file under a function `to_ipaddress` and just call it from both places. If calling ipaddress becomes a bottleneck you can always memoize that function so it's only ever called once per unique string. However, I think this is fine.

supertylerc: I'll leave as-is, then.  Thanks!  :)

yamale/validators/constraints.py
supertylerc: Oops!  Thanks, I'll fix this right away.  There are tests that test for invalid IP addresses that (interestingly) pass without issue.  All tests pass, even after adding these two tests:

```diff
diff --git a/yamale/tests/fixtures/ip.yaml b/yamale/tests/fixtures/ip.yaml
index 3b2d45c..c507a63 100644
--- a/yamale/tests/fixtures/ip.yaml
+++ b/yamale/tests/fixtures/ip.yaml
@@ -3,3 +3,4 @@ ipa: ip()
 ipb: ip()
 ip4: ip(version=4)
 ip6: ip(version=6)
+str: ip(required=False)
diff --git a/yamale/tests/fixtures/ip_bad.yaml b/yamale/tests/fixtures/ip_bad.yaml
index fd86d29..ff7c5f1 100644
--- a/yamale/tests/fixtures/ip_bad.yaml
+++ b/yamale/tests/fixtures/ip_bad.yaml
@@ -3,3 +3,4 @@ ipa: 255.255.255.256/3
 ipb: 2002::dg8/64
 ip4: 2001::db8
 ip6: 192.168.1.1
+str: asdf
diff --git a/yamale/validators/tests/test_validate.py b/yamale/validators/tests/test_validate.py
index 6e53bb7..12a4551 100644
--- a/yamale/validators/tests/test_validate.py
+++ b/yamale/validators/tests/test_validate.py
@@ -113,6 +113,7 @@ def test_ip():
     assert not v.is_valid('192.168.1.256')
     assert not v.is_valid('2001:db8::/129')
     assert not v.is_valid('2001:dg8::/127')
+    assert not v.is_valid('asdf')
 
 def test_mac():
     v = val.Mac()
```

supertylerc: Just pushed up the changes.

supertylerc:
@blopker thanks!  I've updated to address the most recent issue.  Tests still passed the way it was, but I updated it anyway because you are correct.

blopker:
Great! Thanks for sticking with this to the end.

blopker:
Aaaand it's live on PyPi!

PR #51, opened 2019-03-15
Fix cornercase with dots in key name

### LINKED ISSUE #50: Can't Validate Keys with Dots in Them
It seems that I can't validate a YAML file if it has dots (`.`) in the key.  Looks like it's because of the flattening of keys later on, but interestingly you _can_ specify a key with dots in the schema, and Yamale will translate the schema keys with dots in them to underscores.

I worked around this in a _super_ hacky way locally with the following:

```python
from yamale import util

def sub(dic):
    new = {}
    for k, v in util.get_iter(dic):
        if util.is_iter(v):
            v = sub(v)
        if util.isstr(k):
            new[k.replace('.', '_')] = v
    return new


def parse_file(file_name, parser):
    try:
        parse = _parsers[parser.lower()]
    except KeyError:
        raise NameError('Parser "' + parser + '" is not supported\nAvailable parsers are listed below:\nPyYAML\nruamel')
    parsed = parse(file_name)
    new = []
    for x in parsed:
        new.append(sub(x))
    return new
```

This pretty much just replaces `.` with `_` in parsed YAML files, which is being done elsewhere anyway when things get flattened out.  I'm not sure if the flattening functionality that does the dot replacement could be removed with this patch, but maybe it could be.

Would you be open to a cleaned up and tested version of this in a PR?  If so, I can work on it and submit it.

My primary focus is networking, so this is useful when I need to validate a document with IP addresses or MAC addresses as a key.

supertylerc:
Fixes #50.  See that issue for my stream of failures in figuring out how to fix the issue.  I'm not quite sure _exactly_ what causes it to fail, but for some reason there we reach a point where the `.` doesn't get replaced with a `_`.  It only seems to happen when the value of the dotted key is a dict (maybe other non-scalars are affected as well?).


supertylerc:
Knew this would happen.  I'll fix the merge conflicts tomorrow.

supertylerc:
Merge conflicts fixed.

supertylerc:
@blopker, what do you think about this one?  Do you prefer to solve it in another way?

supertylerc:
bump @blopker ?

blopker:
Hey, I'm on vacation right now, I'll check this out soon!

vmiklos:
I just hit this; could you please merge this and tag a new release please? Thanks.

mildebrandt:
Thank you for your interest and your PR. This fix doesn't cover the, very contrived, schema and doc here:
```
items:
  one.two: int()
  one_two: str()
```

```
items:
  one_two: "1"
  one.two: 1
```

This fails with `items.one_two: '1' is not a str`. There may be other cases that are less contrived that would also fall into this category. 

We're going to go in another direction and fix the separator used to flatten the keys, which will avoid the need to change the key names at all during processing.

PR #52, opened 2019-03-20
Support custom YAML tags

rlisbell:
This patch allows PyYAML constructors to be added so that data files may use custom tags.

This affects the interface for `yamale.make_data` but in a backwards compatible manner. There is the potential alternative to handle this with shared variables if that is preferable.

No support for ruamel as I am not familiar with it, but should be easy to add if the library supports custom constructors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ yamale.validate(schema, data)
 
 If `data` is valid, nothing will happen. However, if `data` is invalid Yamale will throw a `ValueError` with a message containing all the invalid nodes.
 
-You can also specifiy an optional `parser` if you'd like to use the `ruamel.yaml` (YAML 1.2 support) instead:
+You can also specify an optional `parser` if you'd like to use the `ruamel.yaml` (YAML 1.2 support) instead:
 ```python
 # Import Yamale and make a schema object, make sure ruamel.yaml is installed already.
 import yamale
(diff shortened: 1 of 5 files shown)


fridex:
Any chance of supporting this feature? It looks handy. Thanks for your response.

mildebrandt:
Hi @fridex, thanks for your interest. This is quite the old PR....before my time here. :) 

Before we look into resurrecting this PR, can you try the solution in this comment?
https://github.com/23andMe/Yamale/issues/145#issuecomment-727116913

jantari:
Hi, I am running into this limitation with ansible variable files containing secrets encrypted with ansible-vault.

[Docs on encrypting YAML keys with ansible-vault](https://docs.ansible.com/ansible/latest/user_guide/vault.html#encrypting-individual-variables-with-ansible-vault)

When attempting to validate these files with yamale, it throws the exception:

    yaml.constructor.ConstructorError: could not determine a constructor for the tag '!vault'

I tried setting the value of the encrypted key to `any()` in the schema because it's not verifiable anyway and I had hoped that `any()` would just make yamale skip over that entry entirely, but the error still occured.

https://github.com/23andMe/Yamale/issues/145#issuecomment-727116913 is not a solution for us because we run the yamale command-line tool in CI, we do not use the python library API so the only configuration options we have is what the CLI flags allow.

We would need a way to either skip encrypted values (or accept them with `any()`), validate them as strings or plug-in a custom validator for them.

mildebrandt:
Hi @jantari, thanks for your interest in Yamale. 

This PR won't allow you to use custom tags from the command line, you'd still need to write your own Python wrapper. It would be best to break out your comment into a separate issue.

However, the way Yamale works isn't compatible with what you're proposing in your schema. Before Yamale validates your yaml file, it will load the file into a Python `dict` using something similar to `yaml.safe_load()`. It's this loader that processes the custom tags and resolves them into strings/ints/arrays/etc. So by the time Yamale begins validation, there are no more strings in the structure that look like `!example`. 

So, we're unable to enforce the existence of yaml tags at this time. You can open an enhancement request under a new issue. 

While I don't have much time to implement new features myself, we do accept pull requests. To get someone started with this support, I think this is a promising way forward: https://stackoverflow.com/a/43768117

And I think the format of `tag('tag_name', validator)` would work nicely. That way we can validate the name of the tag and the value coming into it. For example:

yaml:
```yaml
people:
  - !person
    name: "Bill"
    age: 42
  - !person
    name: "Alice"
    age: 44
```

schema:
```yaml
people: list(tag('person', include(person))
---
person:
  name: str()
  age: int()
```

mildebrandt:

PR #58, opened 2019-07-02
Fix #50 by changing the separator.

### LINKED ISSUE #50: Can't Validate Keys with Dots in Them

mildebrandt:
When flattening keys, the previous separator was a period. This caused
issues when the key itself had a period in it. We changed the separator
to a string that should never appear in the wild.

### DIFF
--- a/yamale/util.py
+++ b/yamale/util.py
@@ -9,6 +9,9 @@ except ImportError:
     from collections import Mapping, Set, Sequence
 
 
+YAMALE_SEP = '`/4M4|_E'
+
+
 # Python 3 has no basestring, lets test it.
 try:
     basestring  # attempt to evaluate basestring
@@ -37,10 +40,8 @@ def flatten(dic, keep_iter=False, position=None):
         return {}
 
     for k, v in get_iter(dic):
-        if isstr(k):
-            k = k.replace('.', '_')
         if position:
-            item_position = '%s.%s' % (position, k)
+            item_position = '%s%s%s' % (position, YAMALE_SEP, k)
         else:
             item_position = '%s' % k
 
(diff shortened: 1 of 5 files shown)


yamale/util.py
```diff
@@ -9,6 +9,9 @@
     from collections import Mapping, Set, Sequence
 
 
+YAMALE_SEP = '`/4M4|_E'
```
blopker: Interesting choice :)

blopker:
Thank you Chris!

vmiklos:
Thanks a lot for fixing this one. Could you please cut a release to enjoy the fix via pip as well?

mildebrandt:
Yes, 1.10.1 is out now.

PR #59, opened 2019-07-02
Fix error message output introduced in PR 58

mildebrandt:
Oops, the separator was used in the error message output, replacing it
with a dot.


blopker:
Nice catch!

Issue #27, opened 2016-03-22
strict mode?


robinpatricksmith:
Keys that do not fit the schema are ignored - so I could have a typo in an optional key and never know about it. It would be nice to have a strict mode where you get a warning that keys were found that are not in the schema.

adrianmo:
+1 for this feature. I also would like Yamale to check keys that are not in the schema.

Thanks for the library anyway!

ericholscher:
+1. This is a primary thing I'd like to have :)

adamf:
+1 this would be great!

milos-korenciak:
+1

alertedsnake:
Is there any interest in doing work on this?  I started taking a crack at it and then realized that it probably means restructuring the whole library.

Just looking to gauge interest from the development team - if it's not a planned feature, I'll be happy to just use a different library.

mildebrandt:
Hi @alertedsnake, thanks for your interest. A little while ago, @drmull mentioned in #60 that they had started working on a PR that would address this. Once that is up for review, I'll be looking at that and #60 to put in together.

mildebrandt:
This is available in version 2.0.

Issue #37, opened 2017-08-01
Adding possibility to include validators


rigal-m:
Hello,
I think it could be interesting to be able to include not only nodes but also validators in Yamale. For example:

```yaml
# [schema.yaml]

key1: include('freq_val_1')
key2: include('freq_val_2')

--- # Frequently used validators
freq_val_1: enum('str_1', 'str_2', 'str_3', 'str_4', 'str_5')
freq_val_2: map(include('a_node_with_some_very_long_name_1'), include('a_node_with_some_very_long_name_2'))
# etc...

--- # Included nodes
a_node_with_some_very_long_name_1:
  n: int()
a_node_with_some_very_long_name_2:
  x: num()
```

I can do a PR if you are interested. In this case all suggestions are welcome.

blopker:
Hey @rigal-m, sorry I'm so slow.

I think that would be a good feature. I'd be happy to field a PR from you!

rigal-m:
OK, I'm working on it.

mildebrandt:
@drmull Is this fixed with your PR?

mildebrandt:
Oh, just saw a comment that says it was fixed. Thanks! This can be found in 2.0.

PR #60, opened 2019-07-03
Handle position with a custom class instead of string

### LINKED ISSUE #56: Unable to validate list of values
Hi,

We trying to validate list of values of spec.initContainers.resources but we are getting following error while validating list of values

**Error:**
```
root@devenv:/media/sf_test_devenv# yamale -s schema.yaml data.yaml
Validating /media/sf_test_devenv/data.yaml...

Error!
Schema: schema.yaml
Data file: /media/sf_test_devenv/data.yaml
  File "/usr/local/lib/python2.7/dist-packages/yamale/command_line.py", line 29, in _validate
    yamale.validate(schema, data)
  File "/usr/local/lib/python2.7/dist-packages/yamale/yamale.py", line 39, in validate
    schema.validate(d)
  File "/usr/local/lib/python2.7/dist-packages/yamale/schema/schema.py", line 56, in validate
    raise ValueError(error_str)
ValueError:
Error validating data /media/sf_test_devenv/data.yaml with schema schema.yaml
        spec.template.spec.initContainers.0.resources.requests.cpu: Required field missing
        spec.template.spec.initContainers.0.resources.requests.memory: Required field missing
        spec.template.spec.initContainers.0.resources.limits.memory: Required field missing
        spec.template.spec.initContainers.0.resources.limits.cpu: Required field missing
        spec.template.spec.initContainers.1.resources.requests.cpu: Required field missing
        spec.template.spec.initContainers.1.resources.requests.memory: Required field missing
        spec.template.spec.initContainers.1.resources.limits.memory: Required field missing
        spec.template.spec.initContainers.1.resources.limits.cpu: Required field missing

  File "/usr/local/bin/yamale", line 11, in <module>
    load_entry_point('yamale==1.10.0', 'console_scripts', 'yamale')()
  File "/usr/local/lib/python2.7/dist-packages/yamale/command_line.py", line 112, in main
    _router(args.path, args.schema, args.cpu_num, args.parser)
  File "/usr/local/lib/python2.7/dist-packages/yamale/command_line.py", line 96, in _router
    _validate_single(root, schema_name, parser)
  File "/usr/local/lib/python2.7/dist-packages/yamale/command_line.py", line 68, in _validate_single
    _validate(s, yaml_path, parser)
  File "/usr/local/lib/python2.7/dist-packages/yamale/command_line.py", line 36, in _validate
    raise ValueError('Validation failed!')
ValueError: Validation failed!
root@devenv:/media/sf_test_devenv#
```

-------------------------
**Data.yaml**
--------------------------
```

##---
# Source: SampleApp/templates/deployment.yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: SampleApp
  labels:
    app: SampleApp
    chart: SampleApp-0.0.0-snapshot
spec:
  replicas: 2
  selector:
    matchLabels:
      app: SampleApp
      release: RELEASE-NAME
  template:
    metadata:
      labels:
        app: SampleApp
        release: RELEASE-NAME
    spec:
      initContainers:
      - name: SampleApp-create-db
        image: "${DOCKER_NAME}/service-db-maria:0.0.18"
        imagePullPolicy: IfNotPresent
        resources:
            limits:
              cpu: 1
              memory: 512Mi
            requests:
              cpu: 50m
              memory: 128Mi
            
      - name: SampleApp-migrate-db
        image: "${DOCKER_REGISTRY}/SampleApp-migrate:${BUILD_VERSION}"
        resources:
            limits:
              cpu: 1
              memory: 512Mi
            requests:
              cpu: 50m
              memory: 128Mi
      containers:
        - name: SampleApp
          image: "${DOCKER_REGISTRY}/SampleApp:0.0.0-snapshot"
          imagePullPolicy: IfNotPresent
          env:
          - name: LOGGING_LEVEL
            value: "info"
          ports:
            - containerPort: 80
            - containerPort: 6060
          resources:
            limits:
              cpu: 1
              memory: 512Mi
            requests:
              cpu: 50m
              memory: 128Mi
```
            
----------------
**Schema.yaml**
----------------
```
spec: 
  template:
    spec: 
      initContainers: list(include('resourcesDefinition'))
---
resourcesDefinition:
  resources: 
    limits:
      memory: any(str(), int())
      cpu: any(str(), int())
    requests:
      cpu: any(str(), int())
      memory: any(str(), int())
```

drmull:
Add datapath class to replace the 'dotted string' used to indicate the path in the data/schema

- Fixes issue with nested maps in an include, closes #56 
- Allows top level validators other than fixed map, brought up in the now closed issue #40


mildebrandt:
Wow, thanks! This initially looks good. I'll need a little more time to go through the whole PR, I'm pretty swamped this week. Can you please fix the PEP8 formatting in the meantime?

yamale/schema/datapath.py
```diff
@@ -0,0 +1,48 @@
...
+
+def _getitems(obj):
+    if isinstance(obj, Mapping):
+        return obj.items()
+    if isinstance(obj, Iterable) and not isstr(obj):
+        return enumerate(obj)
+
+class DataPath(object):
+
+    def __init__(self, key=[]):
```
mildebrandt: Can you modify this to not use a mutable default argument? https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments 

I realize there is no issue in the current code, but I'd like to ensure changes by others don't turn this into a future problem. Thanks!

drmull:
Thanks for the comments! The code actually became a bit simpler when I fixed the mutable default arguments issue..

mildebrandt:
I only looked a little bit at this and haven't taken the time to fully think about it yet. Do you have an understanding how issue #27 would be fixed with this new structure?

drmull:
Yes, I have been working on a follow-up PR that would fix #27 and #37. The idea I have been working on there is to, instead of flattening the path, keep the lists and maps encountered when processing the schema. Then when traversing the schema the list/map keys can be compared with the ones in the data to see that there are no additional keys present.

mildebrandt:
That's great, I look forward to seeing what you come up with!

As you make changes, please run a performance comparison between the original code and the new changes. We run this on tens of thousands of yaml files hundreds of times per day...so performance is important for us. 

And thanks for your contributions!

samnache:
@drmull thanks. really looking forward to this fix.

drmull:
@mildebrandt
Do you want me to update this PR with the strict mode code or push it as a separate PR? I still got some cleanup and documentation left so not 100% yet..
I do not have access to any large enough data files to do a proper performance comparison. Do you have the possibility to help with that?

mildebrandt:
Since it's building on top of this work, it might be easier to review if it's all together. So you can use this PR. 

Sorry, I don't have anything I can share. I can create something for you next week. If you get to it before I do, I think a schema with a few includes and nested lists/maps of 3-4 levels would be sufficient. The total yaml file size should be around 50kB.

drmull:
Finally had some time to do the update..

* Added strict mode to solve issue #27 
* Added possibility to include validators as requested in #37 

The first commit should also fix #62

yamale/schema/schema.py
```diff
@@ -60,77 +63,97 @@ def validate(self, data, data_name):
...
                 return errors
             # SHUT DOWN EVERTYHING
             errors.append('%s: Required field missing' % path)
             return errors
 
-        return self._validate_item(validator, data_item, data, path, includes)
+        return self._validate(validator, data_item, path, strict)
 
-    def _validate_item(self, validator, data_item, data, path, includes):
+    def _validate(self, validator, data, path, strict):
```
drmull: I swapped name of the functions here since I think it makes more sense this way now. Maybe I should also swap place of the functions since validate is the first function called. The diff is nicer this way though..

yamale/schema/schema.py
```diff
@@ -13,44 +12,48 @@ class Schema(object):
     Makes a Schema object from a schema dict.
     Still acts like a dict.
     """
-    def __init__(self, schema_dict, name='', validators=None):
+    def __init__(self, schema_dict, name='', validators=None, includes=None):
```
drmull: I changed it so that included schemas also store the include dictionary in the object instead of passing the include data around since the argument lists where getting a bit long

mildebrandt:
Thanks so much! I'll try to carve out some time to look at this.

yamale/tests/test_functional.py
```diff
@@ -154,15 +191,67 @@ def test_bad_regexes():
     assert count_exception_lines(regexes['schema'], regexes['bad']) == 9
 
 
+def test_bad_include_validator():
+    # TODO use match
```
mildebrandt: What is this TODO referencing?

yamale/tests/test_functional.py
```diff
@@ -154,15 +191,67 @@ def test_bad_regexes():
...
+
+def test_bad_mixed_strict_map():
+    exp = ['field3.extra: Unexpected element']
+    match_exception_lines(mixed_strict_map['schema'],
+                          mixed_strict_map['bad'],
+                          exp,
+                          strict=False)
+
+
+def match_exception_lines(schema, data, expected, strict=False):
```
mildebrandt: I'd rather see this written as the following:
```
def match_exception_lines(schema, data, expected, strict=False):
    with pytest.raises(ValueError) as e:
        assert yamale.validate(schema, data, strict)

    message = str(e.value)
    # only match the actual error message and remove the leading \t
    got = set(s.lstrip() for s in message.split('\n') if s.startswith('\t'))
    expected = set(expected)
    assert got == expected
```

Or if you change the `expected` parameter to a string, you could do this:
```
def match_exception_lines(schema, data, expected, strict=False):
    with pytest.raises(ValueError, match=re.escape(expected)):
        assert yamale.validate(schema, data, strict)
```

mildebrandt:
There's an exception when I try to validate this yaml:
```
field1:
  field1_2:
    field1_2_1: value
field2: value
```
with this schema:
```
field1:
  field1_1:
    field1_1_1: str(required=False)
  field1_2:
    field1_2_1: str(required=False)
field2: str()
```

The error is:
```
Error!
Schema: m1.schema
Data file: /Users/chrism/workspace/drmull/Yamale/m1.yaml
  File "/Users/chrism/workspace/drmull/Yamale/yamale/schema/schema.py", line 75, in _validate_item
    data_item = data[key]
KeyError: 'field1_1'

During handling of the above exception, another exception occurred:

  File "/Users/chrism/workspace/drmull/Yamale/yamale/command_line.py", line 29, in _validate
    yamale.validate(schema, data)
  File "/Users/chrism/workspace/drmull/Yamale/yamale/yamale.py", line 38, in validate
  File "/Users/chrism/workspace/drmull/Yamale/yamale/schema/schema.py", line 56, in validate
    errors = self._validate(self._schema, data, path, strict)
  File "/Users/chrism/workspace/drmull/Yamale/yamale/schema/schema.py", line 98, in _validate
    strict)
  File "/Users/chrism/workspace/drmull/Yamale/yamale/schema/schema.py", line 144, in _validate_static_map_list
    key)
  File "/Users/chrism/workspace/drmull/Yamale/yamale/schema/schema.py", line 84, in _validate_item
    return self._validate(validator, data_item, path, strict)
  File "/Users/chrism/workspace/drmull/Yamale/yamale/schema/schema.py", line 98, in _validate
    strict)
  File "/Users/chrism/workspace/drmull/Yamale/yamale/schema/schema.py", line 144, in _validate_static_map_list
    key)
  File "/Users/chrism/workspace/drmull/Yamale/yamale/schema/schema.py", line 78, in _validate_item
    if validator.is_optional:  # Optional? Who cares.
AttributeError: 'dict' object has no attribute 'is_optional'
```
```

mildebrandt:
Can you please also add a command line argument for strict mode? Thanks!

yamale/tests/test_functional.py
drmull: It was in a reminder for me to update the test case to use the match_exception_lines function. The update was done but I forgot to remove the TODO comment.. Removed now!

yamale/tests/test_functional.py
drmull: Updated to use the first version you suggested. I didn't know the assert would give such a useful error message for set comparison. Nice!

drmull:
Good comments. Fixed the exception problem and added the strict mode flag, also added test cases for both issues.

mildebrandt:
Thanks for all the hard work @drmull! There was one unexpected change which will cause an incompatibility with previous versions, but we think that may be the right thing in the long run.

In the schema and yaml sample I provided in my comments, all the children of `field1_1` were marked as optional which made `field1_1` also optional in the current yamale versions. This is probably a side-affect that we didn't want. Your PR changed this behavior, and now it'll fail unless the parent is also marked as optional.

When we merge your PR, we'll be bumping the version to 2.0 since it breaks existing validations. I don't have a date when we'll merge and create the new release, but we are excited for strict mode....so it won't be too long. :)

drmull:
> In the schema and yaml sample I provided in my comments, all the children of `field1_1` were marked as optional which made `field1_1` also optional in the current yamale versions. This is probably a side-affect that we didn't want. Your PR changed this behavior, and now it'll fail unless the parent is also marked as optional.
> 

Hm, I see. I missed that it worked that way. Spontaneously I think it will be a bit clunky to add that behavior in the code as it looks now so if you are OK with the change as it is then I am happy :+1:

Issue #66, opened 2019-09-09
IndexError when validating list with missing element


drmull:
When using a schema with a static list and trying to validate a list with a missing element there is an uncaught IndexError exception.

Schema:
`[int()]`

Data:
`[]`

mildebrandt:
Thanks, released in 2.0.1!

PR #67, opened 2019-09-09
Fix IndexError when validating list with missing element

drmull:
Fix for #66

When using a schema with a static list and trying to validate a list with a missing element there is an uncaught IndexError exception.


mildebrandt:
Looks great, thanks!

PR #68, opened 2019-09-10
HEL-356: add basic HTML tag validator

23andJim:
I'm going to guess I'm not the first person who's been using Yamale and tried to do this. 

The point is not to validate that all HTML is completely semantically valid, that all attributes are present, and so forth, but to allow the YAML schema author to whitelist HTML tags that are allowed in a simple string field, and to check that they open and close themselves correctly.

### DIFF
--- a/yamale/validators/tests/test_validate.py
+++ b/yamale/validators/tests/test_validate.py
@@ -142,3 +142,23 @@ def test_mac():
     assert not v.is_valid('ab:cd:ef:12:34:56:78')
     assert not v.is_valid('abcdefghijkl')
     assert not v.is_valid('1234567890ax')
+
+def test_html():
+    v = val.Html(tags=['strong', 'em'])
+    assert v.is_valid('<strong>Bold</strong>')
+    assert v.is_valid('<em>Daring</em>')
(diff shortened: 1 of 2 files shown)


yamale/validators/tests/test_validate.py
```diff
@@ -142,3 +142,23 @@ def test_mac():
     assert not v.is_valid('ab:cd:ef:12:34:56:78')
```
blopker: It would also be good to add some YAML tests with actual examples of usage in it.

yamale/validators/validators.py
```diff
@@ -198,6 +202,80 @@ def __init__(self, *args, **kwargs):
...
+    class InvalidTagError(Exception):
+        pass
+
+    class UnclosedTagError(Exception):
+        pass
+
+    def __init__(self, valid_tags, self_closing):
+        HTMLParser.__init__(self)
+        self.valid_tags = valid_tags
+        self.self_closing = self_closing
```
blopker: There's only a handful of 'void-elements' in the HTML spec. You could just hardcode them here and not have people deal with them: https://www.w3.org/TR/2011/WD-html-markup-20110405/syntax.html#syntax-elements

Issue #55, opened 2019-05-13
empty schema generated IndexError exception


mancusoa74:
hello,
assume the simple block of code below:

    schema = yamale.make_schema('schema.yml')

where: 
schema.yml is an empty file

this generates the following exception

File "....../lib/python3.7/site-packages/yamale/yamale.py", line 17, in make_schema
    s = Schema(raw_schemas[0], path, validators=validators)
IndexError: list index out of range

I think would make sense to add a simple check to a non-empty yml schema.

if you think this make sense, let me know if you will accept a pull request for it. I can help to fix it.

The tool is great and I enjoy it

thnx
Antonio

mildebrandt:
Yes, a check for an empty file makes sense in this case. We should not be accepting empty schema files. We'd love a PR for this, let me know if you have any questions. Thanks!

tapaswenipathak:
Hi folks: I added a PR let me know if it helps.

mildebrandt:
Thanks so much! I'm going to be busy this week, so I might not get to it until the week after.

PR #70, opened 2019-09-22
Add a check for schema.yml

tapaswenipathak:
Fixes https://github.com/23andMe/Yamale/issues/55. 

Hi! I like the open source projects of the org, just understanding the culture of the open source contributors in the org by writing a PR. Hope the pr helps. Few more interesting projects have no tickets. :)

### DIFF
--- a/yamale/command_line.py
+++ b/yamale/command_line.py
@@ -20,12 +20,7 @@ schemas = {}
 
 
 def _validate(schema_path, data_path, parser, strict):
-    try:
-        if not os.path.getsize(schema_path):
-            raise ValueError('schema.yml is an empty file!')
-        schema = schemas.get(schema_path)
-    except Exception as e:
-        print(e)
+    schema = schemas.get(schema_path)
     try:
         if not schema:
             schema = yamale.make_schema(schema_path, parser)
--- a/yamale/command_line.py
+++ b/yamale/command_line.py
@@ -21,7 +21,7 @@ schemas = {}
 
 def _validate(schema_path, data_path, parser, strict):
     try:
-        if os.path.getsize(schema_path) > 0
+        if os.path.getsize(schema_path):
             schema = schemas.get(schema_path)
     except Exception as e:
         error = '\nError!\n'
--- a/yamale/command_line.py
+++ b/yamale/command_line.py
@@ -20,7 +20,15 @@ schemas = {}
 
 
 def _validate(schema_path, data_path, parser, strict):
-    schema = schemas.get(schema_path)
+    try:
+        if os.path.getsize(schema_path) > 0
+            schema = schemas.get(schema_path)
+    except Exception as e:
+        error = '\nError!\n'
+        error += 'Schema: %s\n' % schema_path
+        error += traceback.format_exc()
+        print(error)
+        raise ValueError('schema.yml is an empty file!')
     try:
         if not schema:
             schema = yamale.make_schema(schema_path, parser)
--- a/yamale/command_line.py
+++ b/yamale/command_line.py
@@ -21,14 +21,11 @@ schemas = {}
 
 def _validate(schema_path, data_path, parser, strict):
     try:
-        if os.path.getsize(schema_path):
-            schema = schemas.get(schema_path)
+        if not os.path.getsize(schema_path):
+            raise ValueError('schema.yml is an empty file!')
+        schema = schemas.get(schema_path)
     except Exception as e:
-        error = '\nError!\n'
-        error += 'Schema: %s\n' % schema_path
-        error += traceback.format_exc()
-        print(error)
-        raise ValueError('schema.yml is an empty file!')
+        print(e)
     try:
         if not schema:
             schema = yamale.make_schema(schema_path, parser)
--- a/yamale/tests/test_functional.py
+++ b/yamale/tests/test_functional.py
@@ -213,8 +213,7 @@ def test_bad_schema():
 def test_empty_schema():
     with pytest.raises(ValueError) as excinfo:
         yamale.make_schema(get_fixture('empty_schema.yaml'))
-        print(excinfo)
-    assert 'schema.yaml is an empty file!' in str(excinfo.value)
+    assert 'empty_schema.yaml is an empty file!' in str(excinfo.value)
 
 
 def test_list_is_not_a_map():
--- a/yamale/tests/test_functional.py
+++ b/yamale/tests/test_functional.py
@@ -210,6 +210,13 @@ def test_bad_schema():
     assert 'fixtures/bad_schema.yaml' in str(excinfo.value)
 
 
+def test_empty_schema():
+    with pytest.raises(ValueError) as excinfo:
+        yamale.make_schema(get_fixture('empty_schema.yaml'))
+        print(excinfo)
+    assert 'schema.yaml is an empty file!' in str(excinfo.value)
+
+
 def test_list_is_not_a_map():
     exp = [": '[1, 2]' is not a map"]
     match_exception_lines(strict_map['schema'],
--- a/yamale/yamale.py
+++ b/yamale/yamale.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 import sys
+import os
 from .schema import Schema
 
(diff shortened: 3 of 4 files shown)


yamale/command_line.py
```diff
@@ -20,7 +20,15 @@
 
 
 def _validate(schema_path, data_path, parser, strict):
-    schema = schemas.get(schema_path)
+    try:
+        if os.path.getsize(schema_path):
```
mildebrandt: How does this raise an exception if the size of the file is 0?

mildebrandt:
Please add a test for this as well. Thanks!

yamale/command_line.py
tapaswenipathak: `os.path.getsize` - https://docs.python.org/3/library/os.path.html#os.path.getsize

I don't think I get your question. You can find information in the docs. Anything with a `if` clause in Python would check for a non null value. Right? Did I answer your question?

mildebrandt: Your error handling is in an `except` block, but there isn't anything that will raise the exception to get to your error handling code. How did you test that you get the message `schema.yml is an empty file!`?

mildebrandt:
I'd like to see a test for the code you just wrote. You're attempting to detect the case when a schema file is empty and print an appropriate error message. It would be great to see a test for that specific case.

yamale/command_line.py
tapaswenipathak: Does this answer the question - https://repl.it/repls/LightcyanEveryCalculator?

mildebrandt: If you print out the exception message, you'll see that is a `No such file or directory` error. If the file existed and is empty, no exception would be raised.

tapaswenipathak: KO - e8ea91435d73e37f5fafa6efa2241a13d43fa244 should fix.

mildebrandt:
You can run `make test` locally to ensure all the tests pass before you push your changes.

yamale/command_line.py
```diff
@@ -20,7 +20,12 @@
 
 
 def _validate(schema_path, data_path, parser, strict):
-    schema = schemas.get(schema_path)
+    try:
+        if not os.path.getsize(schema_path):
+            raise ValueError('schema.yml is an empty file!')
+        schema = schemas.get(schema_path)
+    except Exception as e:
+        print(e)
```
mildebrandt: People can use Yamale as an API by calling `yamale.make_schema()` directly. Performing this check there may catch more uses than putting the check here.

yamale/yamale.py
```diff
@@ -1,12 +1,20 @@
...
 
 
 def make_schema(path, parser='PyYAML', validators=None):
     # validators = None means use default.
+    # test if the schema.yaml file is empty
+    try:
+        if not os.path.getsize(path):
+            raise ValueError('schema.yaml is an empty file!')
+    except ValueError as e:
+        print(e)
```
mildebrandt: Since we're in the API layer, we shouldn't be printing anything. All you need to do is raise the error.

yamale/yamale.py
```diff
@@ -1,12 +1,20 @@
...
 
 PY2 = sys.version_info[0] == 2
 
 
 def make_schema(path, parser='PyYAML', validators=None):
     # validators = None means use default.
+    # test if the schema.yaml file is empty
+    try:
+        if not os.path.getsize(path):
+            raise ValueError('schema.yaml is an empty file!')
```
mildebrandt: The schema file may not be named `schema.yaml`....it can be anything.

yamale/tests/test_functional.py
```diff
@@ -210,6 +210,13 @@ def test_bad_schema():
     assert 'fixtures/bad_schema.yaml' in str(excinfo.value)
 
 
+def test_empty_schema():
+    with pytest.raises(ValueError) as excinfo:
+        yamale.make_schema(get_fixture('empty_schema.yaml'))
+        print(excinfo)
```
mildebrandt: This print isn't needed.

yamale/tests/test_functional.py
```diff
@@ -210,6 +210,13 @@ def test_bad_schema():
     assert 'fixtures/bad_schema.yaml' in str(excinfo.value)
 
 
+def test_empty_schema():
```
mildebrandt: You should leave the previous test you had which tested the command line flow. Never get rid of valid tests. :)

yamale/yamale.py
tapaswenipathak: Good item.

tapaswenipathak:
Hey @mildebrandt: Done per review.

yamale/yamale.py
```diff
@@ -1,12 +1,20 @@
...
+import os
 from .schema import Schema
 
 PY2 = sys.version_info[0] == 2
 
 
 def make_schema(path, parser='PyYAML', validators=None):
     # validators = None means use default.
+    # test if the schema.yaml file is empty
+    try:
```
mildebrandt: You're catching the exception, wrapping it in another exception, and re-raising it again.

tapaswenipathak: How do you think you can exit out of such an exception?

mildebrandt: I'm going to go on vacation for a few days. Take a little time and think how this can be written better. Then we'll discuss when I get back.

tapaswenipathak: Np, just for getting things of my head:
- https://docs.python.org/3/tutorial/errors.html#handling-exceptions. Statement after a try catch clause is executed. 
- https://doughellmann.com/blog/2009/06/19/python-exception-handling-techniques/ : Best practices suggest you should raise exception or just write a return statement.
- Multiple test I added checks the functionality across that exception.

Happy reading!

mildebrandt: Hi, I'm back. Have you thought about a better way to write the following code?
```
    try:
        if not os.path.getsize(path):
            error_message = os.path.basename(path) + ' is an empty file!'
            raise ValueError(error_message)
    except ValueError as e:
        raise ValueError(e)
```

tapaswenipathak: @mildebrandt: I don't think you have any better option. There is no existing Python predefined function throwing a `ValueError` checking for the empty file. You require the second raise `ValueError` for exiting out of the execution flow. 

\+ review and/or merge ping.

mildebrandt: Hi! Sorry for the delay, I was on another vacation. :) It would be better to just remove the `try` block so it's just the following:
```
if not os.path.getsize(path):
    error_message = os.path.basename(path) + ' is an empty file!'
    raise ValueError(error_message)
```

Let me know if you have any questions about that.

tapaswenipathak: Hello, Just FYI - In any programming language it is bad practice raising exceptions w/o try catch clause and handling that immediately.

yamale/yamale.py
```diff
@@ -1,12 +1,17 @@
...
 import sys
+import os
 from .schema import Schema
 
 PY2 = sys.version_info[0] == 2
 
 
 def make_schema(path, parser='PyYAML', validators=None):
     # validators = None means use default.
+    # test if the schema.yaml file is empty
```
mildebrandt: I've been thinking a little about this. How about checking `raw_schemas` instead of the size of the file. For example:
```python
    if not raw_schemas:
        raise ValueError('{} is an empty file!'.format(path))
```

Or you can even place it within the current try block for something like this:
```python
    try:
        if not raw_schemas:
            raise ValueError('No schema documents found!')
        s = Schema(raw_schemas[0], path, validators=validators)
        # Additional documents contain Includes.
        for raw_schema in raw_schemas[1:]:
            s.add_include(raw_schema)
    except (TypeError, SyntaxError, ValueError) as e:
        error = 'Schema error in file %s\n' % path
        error += str(e)
        if PY2:
            error.encode('utf-8')
        raise SyntaxError(error)
```

tapaswenipathak: I choose first one, so that tests I added doesn't require any further manipulation; lemme know otherwise.

mildebrandt:
Thank you for your contribution!

PR #73, opened 2019-11-11
minor improvement in Regex.__init__

SergBobrovsky:


mildebrandt:
Hi, thanks for your interest in Yamale. Can you describe what improvement this makes? If it's a performance improvement, can you provide some numbers to go along with this? Thanks!

yamale/validators/validators.py
```diff
@@ -153,9 +153,8 @@ class Regex(Validator):
     def __init__(self, *args, **kwargs):
         self.regex_name = kwargs.pop('name', None)
 
-        flags = 0
-        for k, v in util.get_iter(self._regex_flags):
-            flags |= v if kwargs.pop(k, False) else 0
+        flags = sum(v for k, v in util.get_iter(self._regex_flags)
```
drmull: The original code uses a "bitwise or", it is not the same as using sum

mildebrandt: Thanks for watching @drmull :) In this case, it should be fine since they're defined as ints anyway, and the flags won't overlap:
```
SRE_FLAG_TEMPLATE = 1 # template mode (disable backtracking)
SRE_FLAG_IGNORECASE = 2 # case insensitive
SRE_FLAG_LOCALE = 4 # honour system locale
SRE_FLAG_MULTILINE = 8 # treat target as multiline string
SRE_FLAG_DOTALL = 16 # treat target as a single string
SRE_FLAG_UNICODE = 32 # use unicode "locale"
SRE_FLAG_VERBOSE = 64 # ignore whitespace and comments
SRE_FLAG_DEBUG = 128 # debugging
SRE_FLAG_ASCII = 256 # use ascii "locale"
```
But I'm not a fan of change for the sake of change, so I was curious if this solved a problem I wasn't aware of before.

abourree: Prefer bitwise or if same flag can occur many time: SRE_FLAG_TEMPLATE twice become SRE_FLAG_IGNORE_CASE with sum

mildebrandt:
This PR doesn't add any obvious improvements.

PR #80, opened 2020-01-03
replace try:/expect: by 'with pytest.raise(ValueError):'

abourree:
On my Linux box, running test_command_line.py never end.
I search in Google and found nothing.
I disable all test inside and add them one by one, only tests with try/expect: stuck pytest

I Google how to expect exception with pytest: [asserting exceptions with pytest](https://dev.to/wangonya/asserting-exceptions-with-pytest-8hl)

I replace all try:/expect: with pytest.raise() and then:
- excepted test_bad_dir witch I mark skip, none get stuck
- test_empty_schema_file now failed because it didn't raise ValueError


abourree:
The Travis CI build failed only on test_empty_schema_file in test_command_line.py as I explain in my commit.
[Empty schema issue](https://github.com/23andMe/Yamale/issues/55) is not fully corrected: it is not corrected on command line only

abourree:
OK, issue [Empty schema issue](https://github.com/23andMe/Yamale/issues/55) is well fix, test was baddy written, I fix it by adding on data file and run test on it.

abourree:
test_bad_dir test works

yamale/tests/test_command_line.py
```diff
@@ -58,13 +50,13 @@ def test_external_glob_schema(parser):
         os.path.join(dir_path, 'command_line_fixtures/schema_dir/ex*.yaml'), 1, parser)
 
 
-def test_empty_schema_file():
-    try:
+def test_empty_schema_file(capsys):
+    with pytest.raises(ValueError, match='Validation failed!'):
         command_line._router(
-            'yamale/tests/command_line_fixtures/empty_schema',
+            'yamale/tests/command_line_fixtures/empty_schema/data.yaml',
```
mildebrandt: Nice find! This test was missing the final `assert False`. Please add the `data.yaml` file to git. Thanks!

mildebrandt:
> On my Linux box, running test_command_line.py never end.

I'm curious why the tests never completed. How do you run the tests? Can you run `make test` to see if that works with the original code? 

I agree with this PR, I would just like to know why it was hanging. Thanks!

abourree:
With non empty data.yaml that git commit accept as change

abourree:
> > On my Linux box, running test_command_line.py never end.
> 
> I'm curious why the tests never completed. How do you run the tests? Can you run `make test` to see if that works with the original code?
> 
> I agree with this PR, I would just like to know why it was hanging. Thanks!

I'm very beginner in Python
I use pytest from commande line to run tests
I take a try to "make test", after pip install tox, I get:

```
action: GLOB, msg: packaging
cwd: /home/abourree/Projects/Yamale
cmd: /usr/bin/python3 setup.py sdist --formats=zip --dist-dir /home/abourree/Projects/Yamale/.tox/dist
running sdist
running egg_info
error: [Errno 13] Permission denied
```
I continue with pytest

mildebrandt:
Can you try running in an virtual environment:
```
python3 -m venv ./venv
source ./venv/bin/activate
pip install -e .
make test
```

PR #81, opened 2020-01-03
resolves #54: empty data should failed if schema requires something

### LINKED ISSUE #54: empty yml to validate not detected
hello,
assume the simple block of code below (may be not a bug)

    schema = yamale.make_schema('schema.yml')
    cfg = yamale.make_data('config.yml')

    try:
        yamale.validate(schema, cfg)
    except ValueError as e:
        log.critical(e)
        sys.exit(1)

where: 
schema.yml is the yaml schema used to validate a config.yml file

Now let's assume in the schema.yml there a line like:
log_file: str(required=True)

Now if I provide an empty config.yml file, the validation succeed with no exception.
So in this way any config.yml is considered valid.

I think would make sense to add a simple check to a non-empty yml file to validate.

if you think this make sense, let me know if you will accept a pull request for it. I can help to fix it.

The tool is great and I enjoy it

thnx
Antonio

abourree:
resolves #54 by replacing [] in make_data by [(None, path)]


mildebrandt:
Thanks for your PR! I'll take a look.

yamale/tests/test_functional.py
```diff
@@ -127,6 +127,11 @@
     'bad': 'static_list_bad.yaml'
 }
 
+nested_issue_54 = {
```
mildebrandt: I don't see `nested_issue_54.yaml`, can you commit that too?

yamale/tests/test_functional.py
```diff
@@ -176,6 +182,10 @@ def test_bad_nested():
     assert count_exception_lines(nested['schema'], nested['bad']) == 4
 
 
+def test_bad_nested_issue_54():
```
mildebrandt: I'm not a fan of this format of checking the number of lines in the exception message and would like to change the existing tests. For example, you're missing the `nested_issue_54.yaml` file, but the test passes. This means that the "file not found" exception message has the same number of lines as what you're raising.

mildebrandt:
I cloned your repo and couldn't validate your fix. For example:
```
(venv) ~/w/a/Yamale (master|…) $ cat schema.yaml 
log_file: str(required=True)
(venv) ~/w/a/Yamale (master|…) $ cat 54.yaml 
(venv) ~/w/a/Yamale (master|…) $ yamale -s schema.yaml 54.yaml
Validating /Users/chrism/workspace/abourree/Yamale/54.yaml...
Validation success!
```

mildebrandt:
Sorry, I had the wrong branch. I see your validation error is the following:
```
Error validating data /Users/chrism/workspace/abourree/Yamale/54.yaml with schema schema.yaml
         : 'None' is not a map
```

I would expect this instead:
```
Error validating data /Users/chrism/workspace/abourree/Yamale/54.yaml with schema schema.yaml
        log_file: Required field missing
```

abourree:
Better than "log_file: Required field missing" => list all missing filed

yamale/tests/test_functional.py
abourree: I agree that test_functional.py looks complex and just counting errors is not enougth, that why I also did test in test_command_line.py where we can easily check errors

mildebrandt: I'd rather not add more things that I'll have to change later. Can you change this to match the format of the tests below that use `match_exception_lines()`? You can create a separate schema file instead of reusing `nested.yaml` if you want to limit the number of errors in your output. Thanks!

abourree:
Should be ready to merge

mildebrandt:
Awesome, thanks!

Issue #79, opened 2019-12-31
include schema if a reference is valid


abourree:
Hello

I'd like include sub schema if a field is set to true.

Does custom validator can walk into full data tree?
If yes, how to?

my data
```yaml
a: true
b:
  c: 10
```

expecting schema
```yaml
a: bool()
b: includeIf('a', 'isTrue', 'b')
---
isTrue: enum(True)
---
b:
  c: int()
```
Regards,

Arnaud

abourree:
I'm not Python developer, but I take a look to sources.
Validation starts from yamale.validate() which loops over each data entries and validates them against schema.
There is no "global" variable with root data.
It looks complex to validate 'b' if sibling 'a' matches something and I don't speak about more generic implementation to validate 'c' if 'a' from root matches something.

abourree:
I reopen the issue and start implementation on my repo.
I've questions:

To walk in data from a string path, I found two modules to import:
1. [dpath](https://pypi.org/project/dpath/): simple module with simple usage: dpath.util.get(data, 'a/b/c')
2. [YamlPath](https://pypi.org/project/yamlpath/): more powerful but with dependence to ruamel.yaml and python 3.6 minimum
I start with dpath but I'm open to change.

I update setup.py with new module dependency, but that doesn't work with python 2.7:
```
install_requires=['pyyaml','dpath']
```
I don't know how to solved it

mildebrandt:
As discussed in the PR, we've decided that conditional schema based on the content isn't a good fit for Yamale today. Thanks for the suggestion.

PR #82, opened 2020-01-06
New feature: includeIf(<path>, <include to test on path>, <then include> [,<else include>)

abourree:
I propose new keyword includeIf for conditional sub-schema.
The goal is to validate with the following schema:
```yaml
a: bool()
b: includeIf('a', 'isTrue', 'bDefined')
---
isTrue: enum('True')
---
bDefined:
  c: int()
```
This data files where 'b' is not optional but depends of 'a':
```yaml
a: False
```
or
```yaml
a: True
b:
  c: 1
```
Because 'b' could have two formats depends of 'a', the following schema include also else sub-schema:
```yaml
a: bool()
b: includeIf('a', 'isTrue', 'bIfTrue', 'bIfFalse')
---
isTrue: enum('True')
---
bifTrue:
  c: int()
---
bIfFalse: enum('Disabled')
```
validates:
```yaml
a: False
b: Disabled
```
or
```yaml
a: True
b:
  c: 1
```
------------------------
This new feature add dependency on [dpath-python](https://pypi.org/project/dpath/) module.


abourree:
PR will fix #79

mildebrandt:
Thanks for the PR and the detailed README updates. Unfortunately, we don't feel this is a good fit for Yamale right now. We'd like to avoid conditional logic within the schema. Please feel free to fork this project if this feature is something you really need for your work. Let me know if you have any questions. Thanks!

abourree:
Yes, I really need conditional logic, so I will maintains it.

PR #83, opened 2020-01-20
Bugfix/clean output

### LINKED ISSUE #64: Cleaner command line error output
Currently, when yamale is used from the command line and an error is found the output is difficult to parse since schema.validate() will raise and then command_line.main() will also raise.  The raise() from command_line.main() doesn't actually add anything new and the raise() from validate() includes the python stack it makes it hard to find the actual problem - I'm only really interested in what yamale found, not the methods it called to find them.

To deal with this locally, I made some very simple changes that I'd be happy to contribute back if you're interested.  In command_line.main(), I simply wrapped the call to _router():
```python
try:
    _router(args.path, args.schema, args.cpu_num, args.parser, args.strict)
except ValueError:
    # Already processed
    pass
else:
    print('Validation success! 👍')
```
To get rid of the python stack from the output, I changed one line in command_line._validate() from
```error += traceback.format_exc()``` to ```error += str(e)```.

With those changes, I was able to change this output
```
Validating /Users/televi/foo.yaml...

Error!
Schema: bar.yaml
Data file: /Users/televi/foo.yaml
  File "/Users/televi/.venv/lib/python3.6/site-packages/yamale/command_line.py", line 29, in _validate
    yamale.validate(schema, data, strict)
  File "/Users/televi/.venv/lib/python3.6/site-packages/yamale/yamale.py", line 38, in validate
  File "/Users/televi/.venv/lib/python3.6/site-packages/yamale/schema/schema.py", line 65, in validate
    raise ValueError(error_str)
ValueError:
Error validating data /Users/televi/foo.yaml with schema bar.yaml
	borg: 'None' is not a str.

  File "/Users/televi/.venv/lib/python3.6/site-packages/yamale/command_line.py", line 29, in _validate
    yamale.validate(schema, data, strict)
  File "/Users/televi/.venv/lib/python3.6/site-packages/yamale/yamale.py", line 38, in validate
  File "/Users/televi/.venv/lib/python3.6/site-packages/yamale/schema/schema.py", line 65, in validate
    raise ValueError(error_str)
ValueError:
Error validating data /Users/televi/foo.yaml with schema bar.yaml
	borg: 'None' is not a str.

During handling of the above exception, another exception occurred:

  File "/Users/televi/.venv/bin/yamale", line 11, in <module>
    load_entry_point('yamale==2.0', 'console_scripts', 'yamale')()
  File "/Users/televi/.venv/lib/python3.6/site-packages/yamale/command_line.py", line 115, in main
    _router(args.path, args.schema, args.cpu_num, args.parser, args.strict)
  File "/Users/televi/.venv/lib/python3.6/site-packages/yamale/command_line.py", line 97, in _router
    _validate_single(root, schema_name, parser, strict)
  File "/Users/televi/.venv/lib/python3.6/site-packages/yamale/command_line.py", line 68, in _validate_single
    _validate(s, yaml_path, parser, strict)
  File "/Users/televi/.venv/lib/python3.6/site-packages/yamale/command_line.py", line 36, in _validate
    raise ValueError('Validation failed!')
ValueError: Validation failed!
```
to this output
```
Validating /Users/televi/foo.yaml...

Error!
Schema: bar.yaml
Data file: /Users/televi/foo.yaml

Error validating data /Users/televi/foo.yaml with schema bar.yaml
	borg: 'None' is not a str.
```

### PR DESCRIPTION
**Pull request by abourree**
PR should fix #64.
Idea is that:
- yamale.validate(schema, data) returns list of failures an external module can use as API.
- command_line.py
  - if data files are valide, display "Validation Success" and exit 0
  - else, output to stdout list of failures and exit 1
- In case of error like invalid schema format, missing schema file ..., Yamale still raise exception

PR need improvment, on YamaleTestCase for example
Any suggestions are welcome

Arnaud

### CODE REVIEW

**Comment by mildebrandt** [comment]
I'm getting mixed signals here. :) 
```
$ cat 54.schema 
log_file: str(required=True)
$ cat 54.yaml 
$ yamale -s 54.schema 54.yaml
Validating /Users/chrism/workspace/abourree/Yamale/54.yaml...
Error validating data '/Users/chrism/workspace/abourree/Yamale/54.yaml' with '54.schema'
        log_file: Required field missing
Validation success! �
```

**Comment by abourree** [comment]
You're rigth issue #54 is back
BTW test_command_line.py validate route method but not main.
I will fix issue

File: yamale/command_line.py [hunk 1]
```diff
@@ -18,22 +17,15 @@
 
 schemas = {}
 
+status = 0
```
mildebrandt: This doesn't seem to be used.

yamale/schema/schema.py
```diff
@@ -55,14 +56,7 @@ def _parse_schema_item(self, path, expression, validators):
     def validate(self, data, data_name, strict):
         path = DataPath()
         errors = self._validate(self._schema, data, path, strict)
-
```
mildebrandt: It seems there is either a validation error or there isn't. Can you explain why we need a result object rather than just raising an exception here?

yamale/schema/validationresults.py
```diff
@@ -0,0 +1,28 @@
...
+        head_line = "Error validating data '%s' with '%s'\n\t" % (self.data, self.schema)
+        error_str = head_line + '\n\t'.join(self.errors)
+        if PY2:
+            error_str = error_str.encode('utf-8')
+        return error_str
+
+    def isValid(self):
+        return len(self.errors) == 0
+
+    def hasError(self, expected):
```
mildebrandt: This doesn't seem to be used anywhere.

yamale/command_line.py
```diff
@@ -18,22 +17,15 @@
...
-        error += 'Schema: %s\n' % schema_path
-        error += 'Data file: %s\n' % data_path
-        error += traceback.format_exc()
-        print(error)
-        raise ValueError('Validation failed!')
+    if not schema:
+        schema = yamale.make_schema(schema_path, parser)
+        schemas[schema_path] = schema
+    data = yamale.make_data(data_path, parser)
+    return yamale.validate(schema, data, strict)
```
mildebrandt: This is a good change, except for the return....can just let that call raise an exception.

yamale/command_line.py
```diff
@@ -85,18 +77,28 @@ def _validate_dir(root, schema_name, cpus, parser, strict):
 
     print('Found %s yaml files.' % len(res))
     print('Validating...')
+    results = []
```
mildebrandt: Instead of collating and returning results, you could raise an exception with all the errors. Then the `_router` method doesn't have to change and the exception would get passed directly to `main` where it can be handled.

yamale/command_line.py
```diff
@@ -112,8 +114,11 @@ def main():
     parser.add_argument('--strict', action='store_true',
                         help='Enable strict mode, unexpected elements in the data will not be accepted.')
     args = parser.parse_args()
-    _router(args.path, args.schema, args.cpu_num, args.parser, args.strict)
-    print('Validation success! 👍')
+    if _router(args.path, args.schema, args.cpu_num, args.parser, args.strict) == 0:
```
mildebrandt: Here, you could catch the exceptions and print out the result....then you don't need the 0/1 status.

mildebrandt:
Thank you for your PR. I have some concerns that it does a little too much wrapping of the errors. I think the only file that needs to change (other than the tests of course) is the command_line.py. Please let me know if there's something I'm not taking into consideration.

abourree:
Exception or not, that's the goal of this PR.

Cambridge definition of exception:
> someone or something that is not included in a rule, group, or list or that does not behave in the expected way

What is the expected way for a validating tool: answer that input is valid or not.
What is not expected way for a validating tool: missing input files, wrong schema format ...

IMO, trace-back stack out when yaml file is not valid made answer response less understandable.
OK validation failed, but why?
I'd like to understand what's wrong at a glance.

Today, I use Yamale in an Ansible playbook and when validation failed, output becomes absolutely unreadable due to trace-back stack.
This PR may allow for me a new PR: Ansible module.

yamale/command_line.py
abourree: I will remove

yamale/schema/schema.py
abourree: Answer bellow

yamale/schema/validationresults.py
abourree: I've started to write some method we can use in unit test class.
I'm agreed to removed

yamale/command_line.py
abourree: My answer bellow

mildebrandt:
Perhaps I didn't explain myself clear enough, I wasn't suggesting to continue printing the stacktrace to the console.....the whole purpose of this PR is to stop doing that. To avoid misunderstand even more, here are the changes I would expect to solve the issue: https://github.com/23andMe/Yamale/compare/temp/better_output

Please take a look and let me know if you have thoughts on why that wouldn't work. Thanks!

abourree:
I get your point

abourree:
LGTM reports that we should not import yamale in yamale.
We do that to be able to use make_schema, make_data and validate functions.
Why don't we define a Yamale class and then replace "import yamale" by "from yamale import Yamale" ?

mildebrandt:
What is YamaleAssertions?

abourree:
We may want to validate yaml file with yamale in a unit test class which already herit from an AcmeTestCase(TestCase) and don't want to herit from TestCase twice.
Proviging assert functions in YamaleAssertions allow user to select at test class definition:
`class TestMyTest(AcmeTestCase, YamaleAssertions):`
or simply
`class TestMyTest(YamaleTestCase):`

Note that I don't implement yet YamaleTestCase as YamaleTestCase(TestCase, YamaleAssertions).
If you think that is too much changes for this PR, I can rollback.

mildebrandt:
It sounds like that is a bit much. Though I don't see YamaleAssertions in your PR, so I can't really tell. You may have not checked that in.

mildebrandt:
Thanks for all your work. I'm concerned that this PR changes the expected result of the API, which would force a move to a 3.x version. Instead, I'd like to focus on the issue this fixes, which is about making the output cleaner. 

I know you mentioned the following back in January:

> I'd like to removed unnecessary exception and return list of errors up to yamale.validation() which is better for an API than printing out and throwing exception.

But I thought `yamale.validation()` would still raise the same Exception...just suppress the printing. Sorry I misunderstood. 

One possible solution to "have it both ways" is to create a custom exception that extends `ValueError` which contains each result object. If the user catches `ValueError`, the `arg[0]` will be the full text of the error. If the user catches the custom exception, then they can loop through the individual errors.

How does that sound?

abourree:
OK, you don't want to break API, I understand. I will add new exception as you said.

abourree:
I think code update is there.
May be I need to complete doc with new error YamaleError and explain how to catch it.

abourree:
With try/expect samples, I think PR is ready

yamale/command_line.py
```diff
@@ -112,8 +102,12 @@ def main():
     parser.add_argument('--strict', action='store_true',
                         help='Enable strict mode, unexpected elements in the data will not be accepted.')
     args = parser.parse_args()
-    _router(args.path, args.schema, args.cpu_num, args.parser, args.strict)
-    print('Validation success! 👍')
+    try:
+        _router(args.path, args.schema, args.cpu_num, args.parser, args.strict)
+        print('Validation success! 👍')
+    except ValueError as e:
+        print('Validation failed!\n%s' % e.value.message)
```
mildebrandt, line 109: The `ValueError` type doesn't have a `value` attribute. You can probably just use `str(e)` here.

yamale/yamale_error.py
```diff
@@ -0,0 +1,5 @@
+class YamaleError(ValueError):
+    results: []
+    def __init__(self, results):
+        self.results = results
+        self.message = '\n'.join([str(x) for x in results])
```
mildebrandt, line 5: Instead of setting `message` here, you'll want to make a call to the super class to set the string. Otherwise we get something like this when catching and printing out a `ValueError`:
```
Validation failed!
[<yamale.schema.validationresults.ValidationResult object at 0x106078550>]
```

mildebrandt:
Please run `make test` on your local machine. I had thought the build system would run the tests....but that's not happening for some reason. 
 
Thanks!

mildebrandt:
Ah, they are happening...but the PR isn't being updated. You can see them here: https://travis-ci.org/github/23andMe/Yamale/pull_requests

yamale/command_line.py
abourree: It looks that all exceptions have `value` attribute which contains all exception's members
`str(e)` is shorter bu ugly.

yamale/yamale_error.py
abourree: OK, I didn't understand properly how inheritance work in Python, I will correct it.

yamale/command_line.py
mildebrandt: I get the following exception when running this code in Python 3.7.6:
```
Validating /Users/chrism/workspace/abourree/Yamale/2.yaml...
  File "/Users/chrism/workspace/abourree/Yamale/yamale/command_line.py", line 106, in main
    _router(args.path, args.schema, args.cpu_num, args.parser, args.strict)
  File "/Users/chrism/workspace/abourree/Yamale/yamale/command_line.py", line 87, in _router
    _validate_single(root, schema_name, parser, strict)
  File "/Users/chrism/workspace/abourree/Yamale/yamale/command_line.py", line 58, in _validate_single
    _validate(s, yaml_path, parser, strict)
  File "/Users/chrism/workspace/abourree/Yamale/yamale/command_line.py", line 26, in _validate
    return yamale.validate(schema, data, strict)
  File "/Users/chrism/workspace/abourree/Yamale/yamale/yamale.py", line 49, in validate
    raise YamaleError(results)
yamale.yamale_error.YamaleError: [<yamale.schema.validationresults.ValidationResult object at 0x10b714050>]

During handling of the above exception, another exception occurred:

  File "/Users/chrism/workspace/abourree/Yamale/venv/bin/yamale", line 11, in <module>
    load_entry_point('yamale', 'console_scripts', 'yamale')()
  File "/Users/chrism/workspace/abourree/Yamale/yamale/command_line.py", line 109, in main
    print('Validation failed!\n%s' % e.value.message)
AttributeError: 'YamaleError' object has no attribute 'value'
```

abourree: `str(e)` cuts on first \n

mildebrandt: When you fix the inheritance of YamaleError, you should get the full message. For example:
```
$ python
Python 3.7.6 (default, Dec 30 2019, 19:38:28) 
[Clang 11.0.0 (clang-1100.0.33.16)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> e = ValueError('what\nnow')
>>> print(str(e))
what
now
```

abourree:
My Python2.7 env failed to pip dependencies, it looks broken.
Thanks to Travis CI, I understand my error :-)

mildebrandt:
Great! :) 

I'm not sure why Travis stopped updating the PR with the build results.....but it seems to be working again.

abourree:
2 months my PR failed :-(
No it's OK

README.md
```diff
@@ -74,7 +74,28 @@ data = yamale.make_data('./data.yaml')
...
 ```
 
-If `data` is valid, nothing will happen. However, if `data` is invalid Yamale will throw a `ValueError` with a message containing all the invalid nodes.
+If `data` is valid, nothing will happen. However, if `data` is invalid Yamale will throw a `YamaleError` with a message containing all the invalid nodes:
+```python
+try:
+    yamale.validate(schema, data)
+    print('Validation success! 👍')
+except ValueError as e:
+    print('Validation failed!\n%s' % str(e))
```
mildebrandt, line 83: Don't forget to change the code in command_line.py to use `str(e)` too. I'm still getting an error when running from the command line. Here's how I run yamale locally if it helps:
```
python3 -m venv ./venv
. ./venv/bin/activate
pip install -e .
yamale -s my.schema my.yaml
```

mildebrandt:
Yay, this is looking really good. One thing that's not being caught are errors when the schema fails to be parsed. For example:

my.yaml:
```
text: some words
```

schema.yaml:
```
text: mytext()
```

Output:
```
$ yamale -s schema.yaml my.yaml 
Validating /Users/chrism/workspace/abourree/Yamale/test/my.yaml...
  File "/Users/chrism/workspace/abourree/Yamale/yamale/syntax/parser.py", line 16, in parse
    validators)
  File "<ast>", line 1, in <module>
NameError: name 'mytext' is not defined

During handling of the above exception, another exception occurred:

  File "/Users/chrism/workspace/abourree/Yamale/yamale/schema/schema.py", line 50, in _parse_schema_item
    return syntax.parse(expression, validators)
  File "/Users/chrism/workspace/abourree/Yamale/yamale/syntax/parser.py", line 20, in parse
    str(e)
SyntaxError: Invalid schema expression: 'mytext()'. name 'mytext' is not defined

During handling of the above exception, another exception occurred:

  File "/Users/chrism/workspace/abourree/Yamale/yamale/yamale.py", line 19, in make_schema
    s = Schema(raw_schemas[0], path, validators=validators)
  File "/Users/chrism/workspace/abourree/Yamale/yamale/schema/schema.py", line 22, in __init__
    self.validators)
  File "/Users/chrism/workspace/abourree/Yamale/yamale/schema/schema.py", line 41, in _process_schema
    validators)
  File "/Users/chrism/workspace/abourree/Yamale/yamale/schema/schema.py", line 45, in _process_schema
    validators)
  File "/Users/chrism/workspace/abourree/Yamale/yamale/schema/schema.py", line 54, in _parse_schema_item
    raise SyntaxError(error)
SyntaxError: Invalid schema expression: 'mytext()'. name 'mytext' is not defined at node 'text'

During handling of the above exception, another exception occurred:

  File "/Users/chrism/workspace/abourree/Yamale/venv/bin/yamale", line 11, in <module>
    load_entry_point('yamale', 'console_scripts', 'yamale')()
  File "/Users/chrism/workspace/abourree/Yamale/yamale/command_line.py", line 106, in main
    _router(args.path, args.schema, args.cpu_num, args.parser, args.strict)
  File "/Users/chrism/workspace/abourree/Yamale/yamale/command_line.py", line 87, in _router
    _validate_single(root, schema_name, parser, strict)
  File "/Users/chrism/workspace/abourree/Yamale/yamale/command_line.py", line 58, in _validate_single
    _validate(s, yaml_path, parser, strict)
  File "/Users/chrism/workspace/abourree/Yamale/yamale/command_line.py", line 23, in _validate
    schema = yamale.make_schema(schema_path, parser)
  File "/Users/chrism/workspace/abourree/Yamale/yamale/yamale.py", line 28, in make_schema
    raise SyntaxError(error)
SyntaxError: Schema error in file schema.yaml
Invalid schema expression: 'mytext()'. name 'mytext' is not defined at node 'text'
```

I think you should be able to catch `(SyntaxError, NameError, TypeError, ValueError)` in `main()` without changing anything else to cover this case.

mildebrandt:
There seems to be an issue with multiple errors in sub directories. Here's a script that will produce the error:
```
#!/bin/sh
mkdir -p subdir_test/1/1
mkdir -p subdir_test/1/2
mkdir -p subdir_test/2/1
mkdir -p subdir_test/2/2

echo "text: str()" > subdir_test/schema.yaml
echo "text: str()" > subdir_test/1/schema.yaml
echo "text: str()" > subdir_test/2/schema.yaml
echo "text: str()" > subdir_test/1/1/schema.yaml
echo "text: str()" > subdir_test/1/2/schema.yaml
echo "text: str()" > subdir_test/2/1/schema.yaml
echo "text: str()" > subdir_test/2/2/schema.yaml

echo "text: some words" > subdir_test/good.yaml
echo "text: some words" > subdir_test/1/good.yaml
echo "text: some words" > subdir_test/2/good.yaml
echo "text: some words" > subdir_test/1/1/good.yaml
echo "text: some words" > subdir_test/1/2/good.yaml
echo "text: some words" > subdir_test/2/1/good.yaml
echo "text: some words" > subdir_test/2/2/good.yaml

echo "text: 123" > subdir_test/2/2/bad.yaml
echo "text:\n  nested: some words" > subdir_test/1/1/bad.yaml

yamale subdir_test
```

abourree:
No mater if there one or more bad files.
It look that multi treading doesn't like error, specially sharing sdout.
I'm investigating ...

abourree:
Exception in multiprocessing is a nightmare, that why initial implementation which return OK/KO results was good for me.

abourree:
I made half step back:
I add one optional parameter to yamale.validate(): raiseError=True
By default yamale.validate() raise YamaleError if there are errors.
In _validate_dir(), we call yamale.validate(raiseError=False) and raise YamaleError after all file had been processed.
What do you thing about?

mildebrandt:
I think I see why you added the additional parameter to validate()....to prevent the `r.get()` call from raising an exception. Instead of adding the additional parameter, can we place a try/except around the `r.get()` call? It would look more like this:

```python
for r in res:
    try:
        result = r.get(timeout=300)
    except YamaleError as e:
        result = e.results
        areValid = False
    results.extend(result)
```

abourree:
Try first to surround get with try/except.
I also read article explain how to catch error in pool, return it and raise it after.
I failed on both, that why I made this step back.
If you succeed on your side, I will happy.

abourree:
I add parameter to yamale.validate() because I think that's bad to raise in and catch in caller.
IMO optional new parameter in yamale.validate() is acceptable change which not need increase major version.
If you want I can raise error in yamale.validate() and catch it in command_line._validate() if call from validate_dir() and raise it again in validate_dir() and finally catch it in main().
I don't like that, but if you think that better for API compatibility, that's OK for me.

yamale/yamale.py
```diff
@@ -37,7 +38,13 @@ def make_data(path, parser='PyYAML'):
     return [(d, path) for d in raw_data]
 
 
-def validate(schema, data, strict=False):
+def validate(schema, data, strict=False, raiseError=True):
```
mildebrandt, line 41: Please change this to `raise_error`.

mildebrandt:
Ok, I see the issue. The multiprocessing module pickles the exception until it's ready to return it. When it's unpickled, it'll pass all the arguments that were given to the `BaseException` class (which is just a string in YamaleError's case). 

There's one little thing to change, and I think we'll be done with this PR! :) Thanks for sticking with it for this long.

abourree:
Could you provide reference to your coding rules? I don't get signification of underscore prefix.

mildebrandt:
Sure, it's part of the PEP8 style guide: https://pep8.org/#descriptive-naming-styles

```
_single_leading_underscore: weak “internal use” indicator. 
```

It's a hint to other programmers that using this variable should be avoided.

mildebrandt:
I've run this against our set of files....and it misses a few schema errors. It catches them when I run it from one directory deeper. I'll try to get a good recreation. In the meantime this will hold up creating a new release.

There's nothing you need to do, just FYI.

mildebrandt:
FYI: I found some time to fix the missing errors: https://github.com/23andMe/Yamale/pull/101

Should have a new release once that's reviewed and committed.

PR #85, opened 2020-02-03
New schema example added to README.md

europ:
Example describes if the data is a list of items without a key at the top level.

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,6 +434,31 @@ questions:
         - id: 'id_str1'
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -134,6 +134,33 @@ person2:
     age: 20
 ```
 
+If the data is a list of items without a key at the top level:
```
mildebrandt: Thanks for helping our docs. I think this may be better in the `examples` section.

mildebrandt:
Thanks!

PR #86, opened 2020-02-04
Fixed example for 'Custom validators'

europ:
* added missing comma
* added missing module - datetime


mildebrandt:
Thanks for finding that!

PR #93, opened 2020-04-18
Make any() accept anything

salu133445:
Hi, this makes `any()` (without any arguments passed) to accept anything. This actually has the same effect by simply removing the key from the schema file. However, in my use case, I would like the users to see `key: any()` and knows that this key is optional rather than unavailable. Also, things get a bit different in strict mode, which would be helpful when one wants some key to present but does not care about its type. Thanks for reviewing!

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,12 +305,13 @@ Examples:
 
 
 ### Any - `any([validators])`
-Validates against a union of types. Use when a node can contain one of several types. It is valid if at least one of the listed validators is valid.
-- arguments: one or more validators to test values with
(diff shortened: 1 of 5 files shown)


mildebrandt:
That make sense. Can you please update the README with the new functionality?

mildebrandt:
And add some additional tests

salu133445:
Hi, I have updated the README and add some tests. Thanks for reviewing.

mildebrandt:
Looks good, thanks for the PR and thanks for using Yamale!

Issue #62, opened 2019-07-17
Custom key validation


Dejv456:
Hello,

is it possible to validate custom key (**'abc'** or **'bcd'** or whatever **str()**) as below:

**Data:**
```
abc:
  type: 1
  option: a
bcd:
  type: 2
  option: b
```

**Schema:**
```
???: include('profile')
---
profile:
  type: int()
  option: str()
```

mildebrandt:
Hi, there's a PR in review that may allow for this: https://github.com/23andMe/Yamale/pull/60

mildebrandt:
This is available in version 2.0.

Phidica:
Hi @mildebrandt, this may be a dumb question but... Could you please explain *how* to do this now? I couldn't spot this scenario in the readme, and I tried writing the schema like
```yaml
str(): include('profile')
---
profile:
  type: int()
  option: str()
```
but just get
```
Error validating data data.yaml with schema schema.yaml
        str(): Required field missing
```

I tried searching the repo with words like "custom" or "dynamic" but the former is a common word otherwise and the latter is nowhere to be found...
Thanks for your help.

mildebrandt:
Hi @Phidica, thanks for your interest in Yamale. 

For this, it seemed that @Dejv456 didn't want to validate the key per se, but wanted the ability to use anything as the key at the root of the document. To achieve that, we let `map()` be defined at the top level of the schema, like this:
```
map(include('user'))
---
user:
    name: str()
    age: int()
```
Then the following yaml would validate as successful:
```
bob:
    name: Bob Ross
    age: 52
```
If you want actual validation for the keys of a map, that was briefly discussed in issue #53. We would welcome a PR that added the functionality outlined in comment https://github.com/23andMe/Yamale/issues/53#issuecomment-516957631

Phidica:
Ah, okay, thank you for the advice :)

Issue #53, opened 2019-05-06
dict key type validation


milos-korenciak:
Would it be possible to also add validation of key type in dict (={key:val}), please?

I am missing this functionality as almost all validators work along jsonschema principles. There (by definition) is every key string. But not so in YAML (https://yaml.org/spec/1.2/spec.html#id2764196)... The key itself could be almost everything (unique).

(If I am missing something obvious in yamale docs, excuse me in advance.)

mildebrandt:
Hi, thanks for your interest in Yamale! 

Can you describe your use case? Perhaps there's a way to achieve what you need without using a non-scalar key.

milos-korenciak:
I need to have URI as an key. I want to enforce some protocols (thus valiadate the key).
Consider e.g. cache.yaml (!Data in PyYAML stand for "map into object with key !Data"):

`"file:///home/u/myfile.dat" : !Data {"size": 56, "original_url": "file:///home/u/myfile.dat"}
"https://github.com/23andMe/Yamale/issues/53" : !Data {"size": 12345, "original_url": "https://github.com/23andMe/Yamale/issues/53"}`

As you can notice I use that URI key as an "uniqness constraint" over the dict of objects (which contains the key as internal attribute).

mildebrandt:
Just to be sure I understand, you have a yaml with keys as strings and a custom data class as values. And you want the schema to validate that the `original_url` attribute in the data class matches the key of the data class in the yaml? 

If that's the case, how much control do you have with this YAML? My suggestion would be to rewrite the file like this:
```
"file:///home/u/myfile.dat" : {"size": 56}
"https://github.com/123" : {"size": 12345}
```

If you can avoid custom constructors in your YAML, it'll be a lot more portable in the future.

milos-korenciak:
Only I want from Yamale is to **(1)** allow dot (.) in string key + **(2)** check the correctness of URI in that string in regexp way (= it contains "://" somewhere, only allowed char from system charset, etc.).

I do not want Yamale to check the same value of inner attribute and object key at all.
Maybe your way of writing the data is a bit more efficient, but I need transparent and direct mapping from python in-memory objects to YAML with as little code as possible. In RAM this is the only efficient model.

mildebrandt:
Thanks for the additional explanation. Off the top of my head, this may be supported by adding a key constraint to the map validator to do something like this:
```
map(map(str(), int()), key=regex('.*://.*'))
```
[...]

PR #95, opened 2020-04-24
Add support to the "map" validator for constraining the keys

Phidica:
Per discussion in #53 and #62.

Accepting this pull request would add support for a `key =` constraint on the `map` validator that allows another validator to be applied to the keys of the mapping.

I have written some tests, and I am open to writing more test cases if needed. I almost feel like this was *too* simple to implement, so I fear there must be some nasty corner case problems that are yet unsolved.

I will gladly add more commits to update documentation later. 
For now, I am looking for discussion and input on the implementation :)


yamale/validators/base.py
```diff
@@ -62,7 +62,10 @@ def validate(self, value):
         for constraint in self._constraints_inst:
             error = constraint.is_valid(value)
             if error:
-                errors.append(error)
+                if isinstance(error, list):
+                    errors.extend(error)
```
Phidica, line 66: Of course, the `error` object returned by the `Key` constraint test could be a string just like all the other constraints do, rather than a list as I have done. However, I discovered that the tests in the `Key` constraint could be performed in a different order on Python 2.7 vs Python 3.7 (my version) because dictionaries are now guaranteed to remain in insert order, while before they were not. This meant the constructed string could be totally different between different Python versions (because errors may have been encountered in a different order), and this made it difficult to assert the value of this error output in a version-agnostic way. So, I felt a better solution was to extend the total list of errors with each individual error from the `Key` constraint, because this final list is tested in an arbitrary order by the testing suite. It is certainly a sort of workaround to a more fundamental problem though.

mildebrandt:
Awesome, thanks! It may take a little bit for me to get to this, please be patient.

yamale/validators/base.py
mildebrandt: So, @abourree has a solution to that here: https://github.com/23andMe/Yamale/pull/83/files#diff-cbb0754bd27ec59a69abd0f651e48c3bR290

The `match_exception_lines` method was changed to sort the two lists of errors before comparison. I'd rather keep that. I think the best thing to do is to copy over the changes to that method into this PR since #83 is a bigger piece of work and will take longer to get in. 

Sorry, one of you will be getting a merge conflict to resolve depending whose gets in first.

mildebrandt: Oh, I may have misunderstood the issue. The change in #83 is about checking all the errors in multiple validators/constraints and you're concerned with returning multiple errors in your one constraint. This seems fine then, sorry for the confusion.

mildebrandt:
This looks really good. Can you please update the README? Thanks!

Phidica:
@mildebrandt How's this?

mildebrandt:
Thanks for your contribution!

PR #99, opened 2020-05-10
Fix #98

### LINKED ISSUE #98: Python validation versions
In README, we said that Yamale is validated with Python 2.7 and 3.4.
Before PR #81, tox.ini said Python 2.7 and 3.6
Since PR #81, tox.ini said Pythng 2.7 and lastest Python 3

Do we need to validate Yamale on Python 3 older than latest ?
If yes, should we validate from 3.4 to latest ? Or only 3.4 and latest ? Or only 3.4 ?

abourree:
Update to Python 3.6 both tox.ini and README.md
Will fix #98

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ What's YAML? See the current spec [here](http://www.yaml.org/spec/1.2/spec.html)
 Requirements
 ------------
 * Python 2.7+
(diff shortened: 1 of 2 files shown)


mildebrandt:
Thanks! We've decided to drop Python 2.7 support and will be including this change into #107 .

Issue #104, opened 2020-06-25
Ability to specify schema and data without a filename


simonw:
Currently you use the library like this:
```python
import yamale
schema = yamale.make_schema('./schema.yaml')

# Create a Data object
data = yamale.make_data('./data.yaml')

# Validate data against the schema. Throws a ValueError if data is invalid.
yamale.validate(schema, data)
```
It would be really useful if there was an option to provide the schema and data as Python strings instead - that way the could be loaded from a database or from an incoming HTTP request.

Something like this could work

```python
import yamale

schema = yamale.make_schema(content="""
name: str()
age: int(max=200)
height: num()
awesome: bool()
""")

data = yamale.make_data(content="""
name: Bill
age: 26
height: 6.2
awesome: True
""")

yamale.validate(schema, data)
```

mildebrandt:
Thanks for your interest in Yamale! We do accept pull requests if someone is up for it. I like the proposed solution, making the first parameter optional to maintain API compatibility.

simonw:
A couple of implementation challenges: first, it has to work with both Python 2 and Python 3. Secondly, it has to take the two possible parsers into account (pyyaml and ruamel).

Both of those parser libraries operate on a stream passed to their `load_all()` method:

https://github.com/23andMe/Yamale/blob/f5444a3602e07871c4700efe6aaf75c51addb384/yamale/readers/yaml_reader.py#L4-L18

So I think the way to do this is rewrite those two functions to accept a file-like object rather than a filepath on disk. Then rewrite this `parse_file()` method to instead take either a filepath or a string of content:

https://github.com/23andMe/Yamale/blob/f5444a3602e07871c4700efe6aaf75c51addb384/yamale/readers/yaml_reader.py#L27-L32

Proposed new signature (renamed to `parse_yaml()` because it doesn't just do files any more):
```python
def parse_yaml(path=None, content=None, parser="pyyaml"):
    assert (path is none and content is not None) or (path is not None and content is None), \
        "Pass either path= or content=, not both"
    ....
```

simonw:
I can add an extra parametrize decorator here to exercise both the `path=` version and the `content=` version:

https://github.com/23andMe/Yamale/blob/f5444a3602e07871c4700efe6aaf75c51addb384/yamale/readers/tests/test_yaml.py#L26-L35

simonw:
[...]

PR #105, opened 2020-06-25
content= argument for make_schema() and make_data()

simonw:
To allow you to pass in a string of YAML in place of a file path.

Refs https://github.com/23andMe/Yamale/issues/104

### DIFF
--- a/yamale/readers/tests/test_yaml.py
+++ b/yamale/readers/tests/test_yaml.py
@@ -1,3 +1,4 @@
+import io
 import pytest
 from .. import yaml_reader
 from yamale.tests import get_fixture
@@ -9,14 +10,20 @@ KEYWORDS = get_fixture('keywords.yaml')
 
 
 @pytest.mark.parametrize('parser', parsers)
-def test_parse(parser):
-    a = yaml_reader.parse_file(TYPES, parser)[0]
+@pytest.mark.parametrize('use_string', [True, False])
+def test_parse(parser, use_string):
+    if use_string:
+        with io.open(TYPES, encoding='utf-8') as f:
+            content = f.read()
+        a = yaml_reader.parse_yaml(parser=parser, content=content)[0]
+    else:
+        a = yaml_reader.parse_yaml(TYPES, parser)[0]
     assert a['string'] == 'str()'
 
 
 @pytest.mark.parametrize('parser', parsers)
 def test_types(parser):
-    t = yaml_reader.parse_file(TYPES, parser)[0]
+    t = yaml_reader.parse_yaml(TYPES, parser)[0]
     assert t['string'] == 'str()'
     assert t['number'] == 'num()'
     assert t['boolean'] == 'bool()'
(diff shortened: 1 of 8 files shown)


yamale/readers/tests/test_yaml.py
```diff
@@ -9,14 +10,27 @@
...
+    if use_string:
+        with io.open(TYPES, encoding='utf-8') as f:
+            content = f.read()
+        a = yaml_reader.parse_yaml(parser=parser, content=content)[0]
+    else:
+        a = yaml_reader.parse_yaml(TYPES, parser)[0]
     assert a['string'] == 'str()'
 
 
+def test_parse_validates_arguments():
```
mildebrandt, line 24: Thanks for the negative test cases!

mildebrandt:
I wonder if we can fix this error message:
```
yamale.validate(schema, data)
  File "<stdin>", line 1, in <module>
  File "/Users/chrism/workspace/simonw/Yamale/yamale/yamale.py", line 49, in validate
    raise YamaleError(results)
yamale.yamale_error.YamaleError: Error validating data 'None' with 'None'
        age: '26a' is not a int.
```

Since there are no filenames, we get `None` for the data and schema. I think we have three cases:
- No file for schema or data: `Error validating data`
- No file for schema: `Error validating data <filename>`
- No file for data: `Error validating data with schema <filename>`

I'm open to other phrasing too.

simonw:
Good catch, I hadn't seen that. I like those three error messages.

simonw:
OK, implemented those messages. Figuring out how to test it was a bit tricky but I'm happy with the solution I came up with: https://github.com/23andMe/Yamale/blob/ad676dbf9ad10bf7a827ba1c7027e90e10046f05/yamale/tests/test_functional.py#L325-L348

yamale/schema/validationresults.py
```diff
@@ -26,8 +26,13 @@ def __init__(self, data, schema, errors):
...
-        else: 
-            head_line = "Error validating data '%s' with '%s'\n\t" % (self.data, self.schema)
+        else:
+            head_line_bits = ["Error validating data"]
+            if self.data:
+                head_line_bits.append("'{}'".format(self.data))
+            if self.schema:
+                head_line_bits.append("with schema '{}'".format(self.schema))
+            head_line = " ".join(head_line_bits)
             error_str = head_line + '\n\t'.join(self.errors)
```
mildebrandt, line 36: Thanks for the message formatting. Looks like the `\n\t` after the `head_line` was missed in refactoring.

mildebrandt:
I think the rest looks great. We can merge this in after the last bit of message formatting. Thanks for contributing to Yamale!

simonw:
Good catch, fixed.

yamale/schema/validationresults.py
simonw: Fixed in https://github.com/23andMe/Yamale/pull/105/commits/23b49e6d9f807584441eab7ca8b26c867cc1a405#diff-cbb0754bd27ec59a69abd0f651e48c3b

PR #107, opened 2020-06-26
Fix #106: remove Python 2 support

### LINKED ISSUE #106: Remove Python 2 support
Python 2 has been EOL for a few months now and we don't use it internally any more. Let's remove support for Python 2 and cut a new major version of Yamale. We can also enable strict mode be default at the same time.

blopker:
For v3.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ var/
 # IDE dot directories
 .idea/
 .ropeproject/
+.vscode
 
 # Installer logs
 pip-log.txt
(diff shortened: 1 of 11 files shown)


mildebrandt:
Don't forget the README.

blopker:
Will do! What should we add?

blopker:
Oh, I see what you mean.

blopker:
OK, this is ready for review now.

PR #109, opened 2020-07-04
Prevent int and num validators from accepting bool values.

mechie:
Adds a change to the Integer and Numeric validators such that they won't interpret boolean values (False, True) as valid, as well as tests outlining the expected behavior (and confirming that 0 and 1 are still accepted).

This is an issue caused by the origin of Bool in Python, as outlined in [PEP 285 -- Adding a bool type](https://www.python.org/dev/peps/pep-0285/):
> The bool type would be a straightforward subtype (in C) of the int type, and the values False and True would behave like 0 and 1 in most respects (for example, False==0 and True==1 would be true) except repr() and str().

For historical reasons, bool is a subtype of int in Python, but that is not the case in YAML (see e.g. https://yaml.org/type/bool.html). This can cause issues where Yamale validates a YAML document with a bool in an int-only field, and a downstream consumer (whether a non-Python program or a Python program using `str()` such as `jinja2`) ends in an invalid or unexpected state.

The behavior of the current `int` and `num` validators can be emulated by changing instances of `int()` and `num()` to `any(int(), bool())` and `any(num(), bool())`, so there is an easy upgrade path for users relying on this type coercion, if any.

### DIFF
--- a/yamale/tests/fixtures/numeric_bool_coercion.yaml
+++ b/yamale/tests/fixtures/numeric_bool_coercion.yaml
@@ -1,0 +1,2 @@
+integers: list(int())
+numbers: list(num())
--- a/yamale/tests/fixtures/numeric_bool_coercion_bad.yaml
+++ b/yamale/tests/fixtures/numeric_bool_coercion_bad.yaml
@@ -1,0 +1,6 @@
+integers:
+  - False
+  - True
+numbers:
+  - False
+  - True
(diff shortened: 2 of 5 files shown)


mildebrandt:
This looks great, and thanks for the test cases!

mildebrandt:
It will be a little bit until it's in a release since the removal of Python 2 is also in the master branch. It'll be a part of the 3.0 release.

mechie:
Great, thank you for the prompt review/merge!

> It will be a little bit until it's in a release since the removal of Python 2 is also in the master branch. It'll be a part of the 3.0 release.

Makes sense, looking forward to it 👍

PR #112, opened 2020-07-13
Use python_requires in setup.py

haudren:
Current version of Yamale removed support for Python2, however trying to install it on Python2.7 results in an error. Using the python_requires keyword allows user of a specific python version to get the latest compatible package instead.

The error:
```
pip2 install yamale --user --upgrade
Collecting yamale
  Downloading https://files.pythonhosted.org/packages/1e/97/6734831a1316016ae16aa34a1875e77b9e8130f5b400c5bc96c24e56f79f/yamale-3.0.0.tar.gz
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-5_sFTr/yamale/setup.py", line 4, in <module>
        readme = open('README.md', encoding='utf-8').read()
    TypeError: 'encoding' is an invalid keyword argument for this function

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-5_sFTr/yamale/
```

Notes on how to use `python_requires`:
https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

To properly fix this, the 3.0.0 release should unfortunately be pulled down, and replaced by a 3.0.1 release which contains this fix.

Sorry to bother with a Python2 bug, but I hope this can be the last fix.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,7 @@ setup(
     packages=find_packages(),
     include_package_data=True,
     install_requires=['pyyaml'],
+    python_requires='>=3.6',
     entry_points={
         'console_scripts': ['yamale=yamale.command_line:main'],
     },


mechie:
Pulling down 3.0.0 without warning would break any py3 now depending on `yamale=3.0.0`, though 😬 py2's EOL was 2020/01/01, don't think you should risk breaking supported configurations in order to fix expired ones. 

Could tag as 3.0.1 and _eventually_ pull down 3.0.0 after ample warning, and/or add a note to the docs listing the last version supporting py2.7 (e.g. `yamale=<3`).

mildebrandt:
Thanks for bringing this to our attention. It's an unfortunate situation, but I agree with @mechie that 3.0.0 should stay as-is when we release 3.0.1. I've added a note to the 3.0.0 release notes for Python 2 users. 

Thanks for the PR!

haudren:
Thanks for helping out. Unfortunately, we have a few old systems kicking around where a software / hardware mix makes it difficult to move away from Python2.7 at this time. Hopefully, only a few months left :sweat_smile:

haudren:
Sorry to come back to this, but is there any way to somehow re-tag 3.0.0 ? This is causing more headaches than expected on our end. I also realized that if Yamale 3.0.0 is indeed incompatible with Python3.0 ~ 3.5 it will be broken for less ancient systems. I am not a Python packaging expert, but it would be great if we could somehow simply adjust the metadata, without altering the package's content/

mildebrandt:
I don't think that's possible without changing the contents....which would change the hash of the package...which may cause issues with some people who have already moved to the 3.0.0 version. 

As for Python 3.0-3.4, those have reached their end of life. Python 3.5 reaches its end of life in 2 months, so we figured 3.6 was safe to require for the 3.x release.

How does your script/packager/system get Yamale? Can that change to pin the major version to the 2.x release? It's highly recommended to pin all dependencies to at least the major version so that you insulate yourself against breaking changes in APIs.

If using the command line as you have above, you can change it to the following:
```
pip2 install yamale~=2.0
```

This will download the the latest 2.x release (2.2.0 in our case). 

If you're using a requirements.txt file, you can add the following line:
```
yamale~=2.0
```

And if you're using pipenv, add the following line:
```
yamale = "~=2.0"
```

I hope one of those solutions will help. Please pin the rest of your dependencies as well when you get a chance.

mechie:
(Not to derail the conversation about retagging 3.0.0)

In the mean time, on your end, would it be possible to indicate a maximum version in your system's pip requirements?

Python 3.x's pip all support that as far as I can tell (see [PEP 440](https://www.python.org/dev/peps/pep-0440), and [PEP 508](https://www.python.org/dev/peps/pep-0508/) for the spec). Python 2.7.x's docs hint at a more rudimentary version of the same functionality being available (see [here](https://docs.python.org/2/installing/index.html#basic-usage)), so something akin to `python -m pip install "yamale<3.0.0"` or `python -m pip install yamale==2.2.0` might allow you to pull the most recent compatible release automatically.

haudren:
That is an unfortunate circumstance, and thanks, I thought 3.5 would still live for a bit. The main issue here is that the [packager](https://github.com/ros-infrastructure/rosdep) we are using seems to not be compatible with version pinning, as it was really meant to wrap mostly APT... I think all the problems will go away whenever we can upgrade to Python 3 (anytime soon).

Thanks for the advice though!

Issue #114, opened 2020-07-17
add py.typed so that mypy can pick up the typing hints.


cazador481:
pypy needs py.typed so that it knows to pick up the typing hints that are included in the source code.

DavidSouther:
Not completed? There's still no py.typed file in this repo.

PR #115, opened 2020-07-22
Add py.typed so that mypy knows that this package can be typed checked.

cazador481:
This commit contians the py.typed file needed, to satisfy the PEP-561
requirement, so that type checkers know that Yamale can be type checked.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,3 @@
 include *.md LICENSE
 include yamale/tests/fixtures/*.yaml
+include yamale/py.typed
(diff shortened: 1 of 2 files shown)


cazador481:
This is a pull request for issue #114

mildebrandt:
Thanks for your contribution. However, I think there's a little confusion about what py.typed means. The existence of this file indicates that the proper annotations or stubs are available for the package. Yamale has neither, so just adding this file doesn't help with type checking.

To fully fix this, type annotations would need to be added to at least the public APIs....but ideally the whole codebase.

mildebrandt:

PR #122, opened 2020-08-05
Added a UUID validator

YaronFruchtmann:
Added a UUID (v4), a.k.a GUID validator. 
Reference:

1. https://en.wikipedia.org/wiki/Universally_unique_identifier
2. https://tools.ietf.org/html/rfc4122

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,6 +368,12 @@ Validates MAC addresses.
 Examples:
 - `mac()`: Allows any valid MAC address
 
+### UUID - `uuid()`
+Validates a Universally Unique Identifier (UUID).
+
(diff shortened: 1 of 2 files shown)


mildebrandt:
Thanks for your contribution to Yamale. I see this is only for UUID4. It would be great if we could instead validate any UUID version by default. If the user wants to restrict validation to a particular version or versions, we could support the following syntax:

```
uuid(version=[1, 3, 4])
uuid(version=4)
```

Adding a few tests to this would really be appreciated as well. Thanks!

mildebrandt:

PR #126, opened 2020-08-14
Update 'any' with another example and 'firmer' XOR wording.

PeterWhittaker:
The description of *any* as a *union* operator is easily missed, despite the presence of the word 'union': Add emphasis, more direct wording, and a clear, one-type example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,7 +371,7 @@ Examples:
 - `mac()`: Allows any valid MAC address
 
 ### Any - `any([validators])`
-Validates against a union of types. Use when a node **must** contain **one and only** of several types. It is valid
+Validates against a union of types. Use when a node **must** contain **one and only one** of several types. It is valid
 if at least one of the listed validators is valid. If no validators are given, accept any value.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -371,13 +371,15 @@ Examples:
 - `mac()`: Allows any valid MAC address
 
 ### Any - `any([validators])`
-Validates against a union of types. Use when a node can contain one of several types. It is valid
+Validates against a union of types. Use when a node **must** contain **one and only** of several types. It is valid
```
mildebrandt, line 374: Did this mean to say "one and only one"? I could also see this as one of the following:

Use when a node **must** contain **one** of several types.
Use when a node **must** contain **only one** of several types.

PeterWhittaker: Yes, yes it did. I shall fix.

PeterWhittaker: Fixed.

Question: Given that an empty any() allows anything, what happens if none of the specified

mildebrandt:
Thanks for contributing to Yamale!

PR #127, opened 2020-08-14
Added brew/python3/pip3 for Mac OS

PeterWhittaker:
Installation on Mac OS is straightforward, if one has python3, which one can install via brew.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,12 @@ Install
 $ pip install yamale
 ```
 
+*Mac OS users - try this:*
+```
+brew install python3
+pip3 install yamale
+```
+
 ### Manual
 1. Download Yamale from: https://github.com/23andMe/Yamale/archive/master.zip
 2. Unzip somewhere temporary


README.md
```diff
@@ -24,6 +24,12 @@ Install
 $ pip install yamale
 ```
 
+*Mac OS users - try this:*
```
mildebrandt, line 27: Since our requirements already mention Python3.6 above, it would be good to put words to the affect of `If you don't already have Python 3.6 or later installed, try this` or something similar. This will help potentially installing python via `brew` when they already have it installed via `pyenv`.

PeterWhittaker: Hmm, I don't know that it is that clear. I don't have `pyenv`, e.g., I've not had to install it yet.

By default, Mac OS comes with `python 2.7.16`, no `pip`, and as far as I can tell, no `pyenv`. I used the recipe proposed for the `README` and have been having success with `yamale`, including running `tox` tests, and I do not have `pyenv` (just checked with both `python3` and `pip3`.

Might be missing something, though, so any guidance is gladly accepted.

mildebrandt: That process is fine if you don't already have a Python 3 installation on your system. Many people choose not to use brew to manage their Python installations on MacOS. 

Currently, we just tell the user that Python is a requirement. With this update, we would be suggesting a method to install Python. And now that we're explaining how to install Python, we need to put more words around it to ensure they don't overwrite an existing version of Python3. 

More words will also prevent questions such as, "Do I need to install Python via brew for Yamale to work?" It seems silly, but I've seen it. :) 

I'd rather leave Python installation instructions out of this README, but if it helps....we should also ensure it's clear this isn't the only way to install Python 3 and that we don't cause more harm.

PeterWhittaker: Fair points. Brew is pretty conservative, though: If something ships with Mac OS, then brew will, by default, install it in cask-only mode, rather than replace the existing component, e.g., python or openssl - I believe this can be overridden, but that seems fraught with peril to me.

But fair points. There are many ways to get there, favouring one probably isn't a great idea. Let me take another look at the wording. If I cannot think of anything clever, I'll revert to something more generic.

(Semi-relevant aside of potential interest: my .bashrc contains conditions so that when working on Mac OS and if and only if various components of interest are present, various paths are set appropriately; I've been thinking about going one further and having a "stock standard" flag should would allow me to run a shell with none of the casks in my paths....)

PeterWhittaker: Huh, just found this [very interesting page that (tl;dr) suggests pyenv is the way to go](https://opensource.com/article/19/5/python-3-default-mac); I'll look into this myself, that may solve the issue I am having with not being able to `tox` with 3.6.

PeterWhittaker: @mildebrandt OK, I think I have a recipe (I am validating it now):
```
brew install pyenv
pyenv install 3.6 # choose appropriate minor version, e.g.,
pyenv install 3.6.11
pyenv install 3.8 # choose appropriate minor version, e.g.,
pyenv install 3.8.5
cd <where Yamale lives>
pyenv local 3.6.11 3.8.5
pyenv version # for confirmation
```
In my case, the latter command shows
```
3.6.11 (set by /Users/pww/pipelineWork/yamale/.python-version)
3.8.5 (set by /Users/pww/pipelineWork/yamale/.python-version)
```

`pyenv which` shows me what I expected:
```
$ pyenv which python3.6
/Users/pww/.pyenv/versions/3.6.11/bin/python3.6
$ pyenv which python3.8
/Users/pww/.pyenv/versions/3.8.5/bin/python3.8
```
However, `make` fails with `ERROR:  py36: InterpreterNotFound: python3.6`

I'll chase that down a bit, see if I can't figure it out.

PeterWhittaker: Found the missing element, `pip install tox-pyenv`... which complicates things a little, in that one needs to have a native pip to install both `tox` and `tox-pyenv`....

I've got to think through the bootstrapping, because, as you say, not everyone installs python or pip via brew, but you have to have a pip to get tox and tox-pyenv, and you might as well install them only once, right? yeah....

PeterWhittaker: @mildebrandt FWIW, I ended up writing a blog post about this; short answer, [the shortest possible recipe, AFAIK, is here](https://peter-whittaker.com/#just-the-recipe). I can shorten even that for the readme, but I think it may be more than is wanted.

Please let me know! I can withdraw the PR completely or create a new one with content based on that page.

mildebrandt: I think those Python installation instructions would be a little too much to include in the README of our little library. But I'm glad you got a blog post out of it....and thanks for the shout-out of yamale!

PeterWhittaker: No worries, great tool. You should have #130 in your queue now; a different take, higher-level and installation-agnostic. Also mentions tox-pyenv.

PeterWhittaker:
I am going to close and delete this since it has been superseded by #130

PR #128, opened 2020-08-14
Add 'datapath' to 'schema_name' in '_find_schema' to fix #69

### LINKED ISSUE #69: schema file in current directory overrules schemas in deeper directories
When yamale is run on the command line in a file tree of yaml files and schema files, the program wrongly validates against the schema file in the current directory rather than matching the schema files deeper in the file tree.

Example:

```
cwd/
   +- schema.yaml
   +- data.yml
   +- subdir/
       +- schema.yaml
       +- data2.yml
```
When yamale is run from `cwd`, both data.yml and data2.yml are validated against `cwd/schema.yaml`; if the  command is called from another directory (outside or higher in the tree) the file `data2.yml` is validated against `cwd/subdir/schema.yaml`.

The reason is that the glob in `_find_schema` will try to match the schema file name without a path prefix, which will match a schema file in the current directory.

https://github.com/23andMe/Yamale/blob/7a57de7f1cac4284567e8240ca1fe055bc030c1b/yamale/command_line.py#L55

This should probably be changed to match the directory name of the data file plus the name of the schema.

PeterWhittaker:
See comments on #69

### DIFF
--- a/yamale/command_line.py
+++ b/yamale/command_line.py
@@ -51,7 +51,8 @@ def _find_schema(data_path, schema_name):
     """ Checks if `schema_name` is a valid file, if not
     searches in `data_path` for it. """
 
-    path = glob.glob(schema_name)
+    directory = os.path.dirname(data_path)
+    path = glob.glob(os.path.join(directory, schema_name))
     for p in path:
         if os.path.isfile(p):
             return p


yamale/command_line.py
```diff
@@ -51,7 +51,8 @@ def _find_schema(data_path, schema_name):
     """ Checks if `schema_name` is a valid file, if not
     searches in `data_path` for it. """
 
-    path = glob.glob(schema_name)
+    directory = os.path.dirname(data_path)
```
mildebrandt, line 54: This looks great, thanks! Can we add a test for this so it doesn't break again?

PeterWhittaker: Let me poke around with this for a bit, and see what I can do.

PeterWhittaker: I have what I think is a reasonable test; I just need to create a clean branch that has only that on it (there was some mucking about and stashish committing going on to get there). I'll try to get that done over the weekend.

The test is to
1. Create `nestedYaml folder` in `command_line_fixtures`
2. Copy `schema.yaml` from `command_line_fixtures` to `nestedYaml`
3. In `nestedYaml`, create a new `good.yaml` and `schema.yaml`, different from all the rest
4. Append the following to `test_command_line.py`:
```
def test_nested_schema_issue_69():
    curdir = os.getcwd()
    os.chdir('yamale/tests/command_line_fixtures/nestedYaml')
    command_line._router('./','schema.yaml', 1, 'PyYAML')
    os.chdir(curdir)
```

If I understand `router` correctly, this should set the same defaults as if `yamale` was called without arguments: It will validate everything from the current working directory on down.

They are on my rather messy [`myTesting` branch](https://github.com/PeterWhittaker/Yamale/tree/myTesting), if you want to peruse (but ignore some of the other stuff, please).

One note: I do not have 3.6, so my successful `tox` tests ran with 3.8 only.

PeterWhittaker: And a question: some of the tests explicitly set the number of CPUs to 1, others to 4, some use `@pytest.mark.parametrize('parser', parsers)`, others specifically set `PyYAML`....

Is there rhyme or reason to these choices? Should they be consistent? Or does it matter, much?

PeterWhittaker:
@mildebrandt Just wanted to make sure that you saw pull request #129, which replaces this one; I did not want to delete this one until various conversations had finished.

mildebrandt:
Yep! Thanks for double checking.

PeterWhittaker:
Closing this as a dupe of #129

PR #129, opened 2020-08-16
Fix issue #69, overriding of nested schema by parent

PeterWhittaker:
includes test and test data

Replaces [Pull 128](https://github.com/23andMe/Yamale/pull/128)

### DIFF
--- a/yamale/tests/test_command_line.py
+++ b/yamale/tests/test_command_line.py
@@ -92,3 +92,10 @@ def test_bad_issue_54():
     assert 'datetime: Required field missing' in e.value.message
     assert 'nest: Required field missing' in e.value.message
     assert 'list: Required field missing' in e.value.message
+
+def test_nested_schema_issue_69():
+    curdir = os.getcwd()
+    os.chdir('yamale/tests/command_line_fixtures/nestedYaml')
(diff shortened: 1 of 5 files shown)


yamale/tests/test_command_line.py
```diff
@@ -92,3 +92,10 @@ def test_bad_issue_54():
     assert 'datetime: Required field missing' in e.value.message
     assert 'nest: Required field missing' in e.value.message
     assert 'list: Required field missing' in e.value.message
+
+def test_nested_schema_issue_69():
+    curdir = os.getcwd()
+    os.chdir('yamale/tests/command_line_fixtures/nestedYaml')
```
mildebrandt, line 98: Instead of changing the current working directory, you can just refer to the correct path as the first argument in the router. We expand it to the absolute path, so the end result is the same.

PeterWhittaker: Fair enough: I can make that change if that is preferred; I chose the current approach to imitate #69 as closely as possible.

PeterWhittaker: Change made. Merge looks clean.

mildebrandt:
Looks good, thanks for contributing to Yamale!

PR #84, opened 2020-01-27
Simplify manual installation instructions

basnijholt:

PR #135, opened 2020-08-28
Developer install instructions, inspired by #84

PeterWhittaker:
Inspired by PR #84 I've added an `Installing` subsection to the `Development` section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -562,6 +562,19 @@ class TestYaml(YamaleTestCase):
 
 Developers
 ----------
+### Installing
+
+There are several ways to install the latest version of Yamale for development and testing purposes. Please choose your preferred approach:
+
+1. Clone the GitHub repo:
+  - `git clone https://github.com/23andMe/Yamale` or
+  - `git clone git@github.com:23andMe/Yamale.git`
+2. Manually download and install from the latest ZIP archive:
+   1. Download `https://github.com/23andMe/Yamale/archive/master.zip`
+   2. Unzip this to a convenient location and `cd` to that location
+   3. Run `python setup.py install` (may have to prepend `sudo`)
+3. Install directly using `pip`: `pip install https://github.com/23andMe/Yamale/archive/master.zip` (may have to prepend `sudo`)
+
 ### Testing
 Yamale uses [Tox](https://tox.readthedocs.org/en/latest/) to run its tests against multiple Python
 versions. To run tests, first checkout Yamale, install Tox, then run `make test` in Yamale's root


README.md
```diff
@@ -562,6 +562,19 @@ class TestYaml(YamaleTestCase):
...
 Developers
 ----------
+### Installing
+
+There are several ways to install the latest version of Yamale for development and testing purposes. Please choose your preferred approach:
+
+1. Clone the GitHub repo:
+  - `git clone https://github.com/23andMe/Yamale` or
+  - `git clone git@github.com:23andMe/Yamale.git`
+2. Manually download and install from the latest ZIP archive:
```
mildebrandt, line 572: The `Developers` section is meant for people that will be contributing to the Yamale source code. So in this case, option 2 and 3 wouldn't work because they would not have a git environment. I'd be hesitant to even include option 1 because it provides instructions on how to use git, which is not relevant to this README.

PeterWhittaker: Fair points all around. Tempted to delete the pull request, then, as it was intended to address some of the shortcomings of #84, but if #84 is non-development-intended, then how important is the "I don't have pip (or dislike using it) and don't quite have Yamale master and want it" use case?

mildebrandt: Hmm...sorry, I'm not quite sure what direction I'm supposed to be led with your question. :) 

Are you suggesting that we remove the manual installation steps to install master because the use case would be small? https://github.com/23andMe/Yamale#manual

Or are you thinking something else? 

Personally, I can't imagine people wanting/needing to install the master branch....in most cases the latest master is already in a released version.

PeterWhittaker: Yeah, I re-read my comment and I can only make sense of it because I know where my head was when I wrote (low blood sugar deep in a Firebase integration). My apologies for the gigo. Basically, I was wondering whether there was a valid, not-an-edge-case-of-the-worst-kind use case for either 2 or 3, especially 3 (and I was agreeing on your assessment of 1).

My wondering was based on trying to make sense of what sense there was in #84 .

Given that developers are expected to have some git knowledge and given that Yamale is not the place for instructions on git and given that any reasonable Yamale installation is likely to have pip (hard to imagine it without pip), I don't see the use cases...

...hence I do not see the need for this PR...

...nor do I see the need for #84, but that one's not my call*.

My original motivation for this PR was to resolve any lingering documentation issues identified by #84, whether identification was explicit or implicit. Based on our conversation, I am inclined to think that #84 was the result of a very particular and very brief itch being scratched and forgotten.

Unless there is strong objection, I'll close the PR and delete the branch, they were fun to do, but they have no place.

* My sense of orderliness is just strong enough to make the sight of a months-old unresolved PR feel like a hanging thread that should be snipped**.

** Happy snipping!

mildebrandt: Ha, thanks! I agree, I'll go ahead and close #84 as well so it doesn't confuse others. 

Have a great weekend.

PeterWhittaker:
Closing as per discussion above.

Issue #133, opened 2020-08-24
"map" validator doesn't validate key, when using include


ivocavalcante:
First of all, congrats for the great job you've done. Yamale is exactly I was looking for a project I'm working on. Now, to the bug report:

"map" validator fails to validate a key, when function used is "included". See example below:

Consider YAML data below, for testing purposes:

**test.yaml**
```
maptest: 
  1: xpto 
  2: qwerty 
  error: wrong
```

**WORKING** (meaning _validation fails_):

**schema.yaml**
```
maptest: map(include('value'), key=int()) 
--- 
value: str()
```

**NON-WORKING** (meaning _validation passes_):

**schema.yaml**
```
maptest: map(include('value'), key=include('key')) 
--- 
value: str()
key: int()
```

Tested with latest release, 3.0.2 .

mildebrandt:
As mentioned in the declined PR, the best course of action is to prevent the `include` validator from being defined in the `key` constraint for the `map` validator. Since this is a potentially breaking change, we should wait for a major release.

martinvonwittich:
Hm, I had hoped to use `include` here like this:

<details>
<summary>schema:</summary>

```
---
users: map(include("user"), key=include("act"))
version: enum("1.0")

---
user:
    firstname: str()
    lastname: str()
    roles: list(include("role"))
    groups: list(include("act"))
    password: include("password")
    samba_password: include("password")
    admin_password: include("password")
    maintain_groups: include("maintain_memberships")
    maintain_roles: include("maintain_memberships")
    maintain_password: include("maintain_password")

---
maintain_memberships: enum("initial_only", "restore_memberships", "fixed")

---
maintain_password: enum("initial_only", "fixed")

---
# /usr/share/perl5/IServ/Valid.pm sub Role
role: str(matches="^ROLE_[A-Z0-9_]{1,64}$")

---
# /usr/share/perl5/IServ/Valid.pm sub Act + MAX_LEN
act: str(matches="^[a-z][a-z0-9._-]{0,31}$")

---
password:
  type: enum("plain", "hashed")
  value: str()
```
</details>

<details>
<summary>input:</summary>

```
---
version: "1.0"
users:
  martintest:
    firstname: Martin
    lastname: Test
    groups:
      - admins
      - domain.admins
    roles:
      - ROLE_ADMIN
    password:
      type: plain
      value: foobar
    samba_password:
      type: hashed
      value: foobar
    admin_password:
      type: plain
      value: barfoo
[...]

PR #136, opened 2020-08-28
Add new base validator and base constraint classes

drmull:
Add new base classes to validator and constraint that take additional arguments, such as schema, path etc. This allows constraints to use non-primitive validators and resolves issue #133 

My initial intention was to pull out the logic for the non-primitive validators from the schema file and put them in each respective validator class but then I figured it would be a bit too drastic..

### DIFF
--- a/yamale/validators/base.py
+++ b/yamale/validators/base.py
@@ -1,4 +1,4 @@
-class Validator(object):
+class BaseValidator(object):
     """Base class for all Validators"""
     constraints = []
     value_type = None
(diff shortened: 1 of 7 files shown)


yamale/validators/base.py
```diff
@@ -30,37 +30,32 @@ def tag(self):
...
-        valid = self._is_valid(value)
-        if not valid:
-            errors.append(self.fail(value))
+        # Optional field with optional value? Who cares.
+        if (value is None and
+                not self.is_required and
+                self._value_can_be_none):
+            return []
+
+        errors = self._validate(value, schema, path, strict)
```
drmull, line 52: I forgot to change the name of the _validate function, I will fix it

yamale/validators/constraints.py
```diff
@@ -107,20 +119,22 @@ def _fail(self, value):
...
-class Key(Constraint):
+class Key(BaseConstraint):
     keywords = {'key': Validator}
     fail = 'Key error - %s'
 
-    def _is_valid(self, value):
+    def _is_valid(self, value, schema, path, strict):
         for k in value.keys():
-            if self.key.validate(k) != []:
+            errors = schema._validate(self.key, k, DataPath(), strict)
```
drmull, line 128: I forgot to change the name of the _validate function, I will fix it

mildebrandt:
Welcome back @drmull :) 

FYI: This may take me a while to get to, it's a busy time over here. Sorry.

drmull:
> Welcome back @drmull :)

Thanks!

> 
> FYI: This may take me a while to get to, it's a busy time over here. Sorry.

No problem, I am in no rush :)

mildebrandt:
Sorry for the long delay. 

I grabbed the example from #133, but I couldn't get your code to work with it. 

**data**:
```
maptest: 
  1: xpto 
  2: qwerty 
  error: wrong
```

**schema**:
```
maptest: map(include('value'), key=include('key')) 
--- 
value: str()
key: int()
```

This combination should fail validation.

drmull:
Hm, strange. I updated the test case to match the example but I get:

`maptest: Key error - 'error' is not a int.`

Do you use the PyYAML parser? Maybe the order of the elements can differ with different parsers and trigger the problem.

I will take a closer look!

mildebrandt:
Yeah, I'm using PyYaml 5.3.1.

However, I'm beginning to think this isn't something we should support. Allowing for `include` in the `key` constraint only makes sense when processing complex keys, for example:

```
? {one: 1, two: 2}
: [ 2001-01-01, 2002-02-02 ]
```

Both PyYAML and ruamel fail to load such a document, but it's valid yaml. Instead, I think the correct course of action is to raise an error saying something like: 

```
The include validator is not valid for the key constraint.
```

Other validators can be used to constrain the `key` and still support all yaml documents that PyYAML and ruamel can load.

drmull:
Yeah, I can agree with that. I will close this PR and we can reopen it if ruamel implements complex keys in the future :grin:

mildebrandt:
Cool, and thanks for your willingness to always jump in!

PR #149, opened 2020-12-15
[Feature] Adds Subset validator

idantene:
Similiar to the `any` validator, except it allows combining multiple types (instead of exactly one).
This is mostly useful for various `Include`s, thereby allowing some sort of conditional key-value pairs, such as the following. Granted this example is over-simplistic and can be resolved in other ways (such as explicitly specifying the mapping values), but this would apply to a recursive definition as well.
```yaml
# schema.yaml
my_types: union(include('type1'), include('type2'))
---
type1: map(str(), key=regex('^my_str(ing)?$', ignore_case=True))
---
type2: map(int(), key=regex('^my_int(eger)?$', ignore_case=True))

# data_success.yaml
my_types:
  my_string: foo
  my_int: 3

# data_fail.yaml
my_types:
  my_string: 3  # Not a string
  my_int: foo  # Not an int
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -389,6 +389,19 @@ Examples:
 - `any(str(min=3, max=3),str(min=5, max=5),str(min=7, max=7))`: validates to a string that is exactly 3, 5, or 7 characters long
 - `any()`: Allows any value.
 
+### Subset - `subset([validators], allow_empty=False)`
+Validates against a subset of types. Unlike the `Any` validator, this validators allows **one or more** of several types.
+As such, it *automatically validates against a list*. It is valid if all values can be validated against at least one
+validator.
+- arguments: validators to test with (at least one; if none is given, a `ValueError` exception will be raised)
+- keywords:
+    - `allow_empty`: Allow the subset to be empty (and is, therefore, also optional). This overrides the `required`
+flag.
+      
+Examples:
+- `subset(int(), str())`: Validators against an integer, a string, or a list of either.
+- `subset(int(), str(), allow_empty=True)`: Same as above, but allows the empty set and makes the subset optional.
+
 ### Include - `include(include_name)`
 Validates included structures. Must supply the name of a valid include.
 - arguments: single name of a defined include, surrounded by quotes.
--- a/yamale/schema/schema.py
+++ b/yamale/schema/schema.py
@@ -110,8 +110,8 @@ class Schema(object):
         elif isinstance(validator, val.Any):
             errors += self._validate_any(validator, data, path, strict)
 
-        elif isinstance(validator, val.Union):
-            errors += self._validate_union(validator, data, path, strict)
+        elif isinstance(validator, val.Subset):
+            errors += self._validate_subset(validator, data, path, strict)
 
         return errors
 
@@ -189,7 +189,7 @@ class Schema(object):
 
         return errors
 
-    def _validate_union(self, validator, data, path, strict):
+    def _validate_subset(self, validator, data, path, strict):
         def _internal_validate(internal_data):
             sub_errors = []
             for val in validator.validators:
--- a/yamale/schema/schema.py
+++ b/yamale/schema/schema.py
@@ -205,16 +205,15 @@ class Schema(object):
             return []
 
         errors = []
-        if isinstance(data, dict):
+        if util.is_map(data):
             for k, v in data.items():
                 errors += _internal_validate({k: v})
-        elif isinstance(data, list):
+        elif util.is_list(data):
             for k in data:
                 errors += _internal_validate(k)
         else:
             errors += _internal_validate(data)
         return errors
-
 
     def _validate_primitive(self, validator, data, path):
         errors = validator.validate(data)
--- a/yamale/schema/schema.py
+++ b/yamale/schema/schema.py
@@ -110,6 +110,9 @@ class Schema(object):
         elif isinstance(validator, val.Any):
             errors += self._validate_any(validator, data, path, strict)
 
+        elif isinstance(validator, val.Union):
+            errors += self._validate_union(validator, data, path, strict)
+
         return errors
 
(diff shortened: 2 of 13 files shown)


mildebrandt:
Hi, thanks for your interest in Yamale.

Is this a feature that you'd like to use in Yamale? Do you have a specific yaml in mind that you'd like to validate that you can't with the existing validators?

As for your code, I'm curious how it would work in certain scenarios. Using your sample schema, the following would be valid:

```
my_types:
  my_string: foo
```

I would think that both maps are required since neither have the `required=False` parameter.

Also, what if the schema was like this:

```
my_types: union(include('type1'), include('type2'))
---
type1: map(str(), key=regex('^my_str(ing)?$', ignore_case=True))
---
type2: int()
```

What would be the expected behavior? The current code would say this is valid:
```
my_types: 3
```

but I don't think that's what you wanted.

I see what you're trying to do, but I don't see how to write this in a way that ensures a standard way of validation. Do you have any thoughts?

idantene:
Hey @mildebrandt, thanks for your quick reply!
It is a feature I'd like to use, but perhaps one that requires more information/examples.

Indeed the `required` keyword is redundant in this case. If the `any` type is meant as XOR, then this `union` type is meant as OR. I haven't seen the `required` keyword used in `any(include(...), include(...))` either, but perhaps it's an oversight on my side. With that, yes, your first example is valid. The second example is valid, but probably not a good use case/example.

Here's my real-life use case - I have a list of self-defined `filters` that are internally translated to SQL queries of various types. This `filters` keyword itself is optional, but if provided, there are various possible combinations of values:
- It could have a simple string key and integer/string value, implying the column in the key should equal the value.
- There are comparison keywords (e.g. `$lt`, `$eq`, `$in`, etc), which accept a similar mapping (with an optional list of value for some of these).
- There are operator keyword (e.g. `$or`, `$and`, etc), which accepts a mapping of the comparison keywords or the operator keywords (but must end in a proper mapping).
- There are top-level only keywords (e.g. `$distinct`), which accept a list of string (column names).

As a concrete example, this is valid:
```yaml
filters:
  $distinct:  # Can only exist in the top-level mapping, requires list of strings
    - foo
    - bar
  baz: 1  # Can exist anywhere, implies `$eq: { baz: 1 }`
  $lt:
    my_value: 7
  $or:
    $eq:
      my_value: 2
    $in:
      my_value: [1, 5, 6, 7]
```

I couldn't find a way to ensure these all live in harmony with their specific requirements otherwise.

mildebrandt:
I see, that is a tricky format. Can you provide a schema for that example that you think will work using your `union()` validator? I still think there's something missing and I'd like to flush it out a bit more. Thanks!

idantene:
I thought something such as the following will work (or rather, it does work but I haven't tested all edge cases):
```yaml
# Comparison filters should have a strict mapping
comp_filters: map(map(str(), int()), key=regex('^\$(eq|lte?|leq|gte?|geq|neq?|i?like|n?in)$', ignore_case=True))
---
# Operational filters may either consist of other operational filters, or have a nested comparison filters
op_filters: map(include('op_filters'), include('comp_filters'), key=regex('^\$(or|and|not)$$', ignore_case=True))
---
select_filters: map(list(str()), key=regex('^\$(distinct)$', ignore_case=True))
---
# Top level filters can either be:
filters: union(include('op_filters'),  # Operational filters, such as $and, $or, $not
               include('comp_filters'),  # Comparison ones
               include('select_filters'),  # The select filters
               map(str(), int(), key=str()))  # Or a direct mapping, implying $eq
```

idantene:
Pinging about this @mildebrandt, any follow-up thoughts?

mildebrandt:
Thanks for the poke! Sorry it's taking a long time to get to this, we're pretty busy over here. Hopefully I can schedule some time soon.

mildebrandt:
I think we can include this functionality. A couple things:

1. I'm not so keep on the name. A union of things means that it includes everything. This is more of a "one or more" scenario. I wish we didn't already have `any()` defined the way it is. I don't have a good name in mind yet. We can either find a new name, or we can add a keyword to the `any()` validator to accept more than one match....behaving like your `union()` validator. What do you think?
2. Of course, tests, tests, tests. :) 

Thanks!

idantene:
I agree the `union` tag is misleading. I don't mind going down the `any` flag with an appropriate flag (how would you name it?). Another option would be perhaps to rename it as `subset`? I think that captures the intent (adding `required=False` allows the empty subset).

I'll see about adding tests in the coming week :+1:

mildebrandt:
I'm good with using `subset()` instead of `union()`. I would probably use `allow_empty` instead of `required`. My first thought with `required` would be that all items are required to be present when it's `True`.

idantene:
Hey @mildebrandt! Sorry it took me this long to get back to this.
I've added a test suite and completed the functionality (and renaming as suggested). I've also added the notes and some basic examples to the README file.

Some notes (for your review process):
- I've made the `test_good` a bit more flexible by allowing more keys with e.g. `good2`, `good_with_X`, etc.
- The `allow_empty` of course, as suggested, overrides both the `required` and `none` flags that work with other validators. I'm not sure if this is worth mentioning in the README somewhere, but I've taken the liberty to explain it a bit.
- The `subset` requires at least one validator (otherwise it doesn't make much sense? or it could imply `allow_empty` to be `True`, but I think users should be explicit about their intent).
- Of course, the `subset` validator suggests a list is _possible_ (not mandatory). I've mentioned this in the README as well.

mildebrandt:
Thanks for the updates! Please be patient....it may take a bit for me to get to the review.

idantene:
Take your time :relieved: I just had this in my to-do list for the weekend.

mildebrandt:
Almost 3 months later! My apologies. This looks good, I'll create a new release for it. Thanks for the contribution!

PR #160, opened 2021-06-14
Add LengthMin and LengthMax constraint to Map

JesseHines0:
Add the ability to validate the size of a map with `min` and `max` like Yamale already can with lists and strings. This can be particularly useful for validating a non-empty map.


yamale/tests/fixtures/map.yaml
```diff
@@ -1 +1,3 @@
 map: map(str(), int())
+min: map(int(), min = 1)
+max: map(int(), max = 2)
```
mildebrandt, line 3: Can you add a test which includes both min and max?

mildebrandt:
This looks good, thanks for the PR. If you can implement that extra test, I'll merge it in. Thanks!

yamale/tests/fixtures/map.yaml
JesseHines0: Done!

mildebrandt:
Well, it seems that travis-ci.org is closed for business as of today. Guess I should have looked at this yesterday. :) 

I never got a notification. It'll take a bit to get up and running with another solution....please be patient, it may be a while.

JesseHines0:
Alrighty, thanks!

mildebrandt:
We got our brand new github actions working, and your addition has made it into version 3.0.7. Thanks for your contribution!

PR #161, opened 2021-06-16
Replace TravisCI with Github Actions

scriptsrc:
Adding:
- `run-tests.yml`: Executes tox for py3.6 and py3.8
- `pypi-publish.yml`: Uses pypi token and uploads new version to pypi. *NOTE*: We'll need to upload the token as a secret.

Removing:
- `.travis.yml`

Modified:
- `README.md`: Points to GH Action Badge instead of Travis Build badge.
- `setup.py`: Updated the list of python classifiers. It was showing py3.4
- `requirements.txt`: Removed `tox-travis`
- `tox.ini`: Added `[gh-actions]` section for `tox-gh-actions`. (Replacement for `tox-travis`)

I'm unsure why Actions wasn't accepting my `$default-branch` variable, so it's hardcoded in both yml files to `master`.


.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,32 @@
+# This workflows will upload a Python Package using Twine when a release is created
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+
+name: Upload Python Package
+
+on:
```
mildebrandt, line 6: The travis script uploaded new versions on tag events. Can github actions do that?

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,34 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Run Tox
+
+on:
+  push:
```
mildebrandt, line 7: We'd like builds to happen on every branch, not just master.

scriptsrc: Fixed in 3c18ad8 (next commit)

.github/workflows/pypi-publish.yml
scriptsrc: Reading about how actions parses that now:

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

sarahc23: This sounds like what you're looking for? https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

```
on:
  push:
    # Sequence of patterns matched against refs/heads
    branches:    
      # Push events on main branch
      - main
      # Push events to branches matching refs/heads/mona/octocat
      - 'mona/octocat'
      # Push events to branches matching refs/heads/releases/10
      - 'releases/**'
    # Sequence of patterns matched against refs/tags
    tags:        
      - v1             # Push events to v1 tag
      - v1.*           # Push events to v1.0, v1.1, and v1.9 tags
```

scriptsrc: They do have an on-release trigger:

```
# Trigger the workflow on pull request activity
on:
  release:
    # Only use the types keyword to narrow down the activity types that will trigger your workflow.
    types: [published, created, edited]
```

Github Releases are based on tags.  Have you been manually creating these releases or does Github auto-create these when you push new tags?

https://github.com/23andMe/Yamale/releases

Alternatively, we can also trigger on tag push/creation:

```
on:
  push:
    # Sequence of patterns matched against refs/heads
    branches:    
      # Push events on main branch
      - main
    # Sequence of patterns matched against refs/tags
    tags:        
      - v1             # Push events to v1 tag
      - v1.*           # Push events to v1.0, v1.1, and v1.9 tags
```

or

```
on:
  push:
    tags:
    - '*'
```

mildebrandt: I think that last example looks like what we'd want. The releases are created by hand, usually only when something very interesting goes in. We'd still want to push to pypi for every new tag that's created.

.github/workflows/run-tests.yml
mildebrandt: Should the actions have run when you pushed this change? Where would the output of the actions be seen?

.github/workflows/pypi-publish.yml
scriptsrc: Updated to run on push for tags

.github/workflows/run-tests.yml
scriptsrc: You can see it running against my fork here:
https://github.com/scriptsrc/Yamale/actions

But these actions won't run on 23andMe's copy until we merge these yml files to master.

mildebrandt (approved):

PR #165, opened 2021-08-03
Limit the strings which can be eval'd

mildebrandt:
When processing the schema, each line is run through Python's eval
function to make the validator available. A well constructed string
within the schema rules can execute system commands. This change limits
the string to those that begin with the known validators.


cblakkan (approved):

hashim23andme (approved):

seungwooc-23 (approved):

gauravphoenix:
Thank you for finding and fixing it, are you planning to publish a CVE for this?

mildebrandt:
Yes, it's CVE-2021-38305

PR #171, opened 2021-09-30
Adding --must-exist command line option

jtyr:
This PR is adding `--must-exist` command line option that allows to exit the execution with an error if the `PATH` doesn't exist. Without this option, the execution seems to be successful which might be a problem if we want to validate schema of a particular file/directory and the file/directory doesn't exist.


mildebrandt:
I think this should be the default behavior, but someone is probably depending on it. This solution seems fine, and I'd like to make it the default in the next major release. Please add tests, thanks!

cblakkan:
Will make this default behavior soon #261

PR #190, opened 2022-02-06
Make file-encoding UTF-8 by default and let users choose

### LINKED ISSUE #189: File encoding is always the system one and cannot be changed
In this line, `open()` is used without any encoding parameter, so the system one is always chosen: https://github.com/23andMe/Yamale/blob/master/yamale/readers/yaml_reader.py#L34

This leads to users not being able to read UTF-8 encoded files on Windows.

It should be possible to set the encoding as an optional parameter during `make_schema()` and `make_data()`.

I would advocate for the encoding being UTF-8 by default.

reconman:
Fixes #189 

I've added some type hints and comments to the user-facing functions. Wasn't sure about the type of `validator`, so it's still `Any`.


mechie:
As an aside, I think this would be a breaking change--Windows-only users (or anyone with [a supported non-utf-8 text encoding](https://docs.python.org/3.7/library/codecs.html#standard-encodings) set as their locale) may be unwittingly, or intentionally, relying on this behavior.

mildebrandt:
I agree @mechie , more comments in #189 .

reconman:
I've changed it to the system default. Tested it with the value `Chūnibyō demo Koi ga Shitai!` in a YAML file. If I encode it as CP1252, it can be parsed on Windows. If I use UTF-8 instead, an exception is thrown.

Not really happy with copy-pasting the locale function call everywhere but I'm not a professional Python programmer, so there may be better ways.

PR #191, opened 2022-02-10
Check for UnicodeEncodeError when printing the success message

sirykd:
Hi,

Thanks for your great work!

There exists an issue I've encountered in several environments with printing "👍".
The problem is that we get "Validation failed!" trying to print "Validation success! 👍" which is both funny and incorrect.

The issue can be reproduced in ubuntu docker, but it is not limited to docker in my experience.

I prefer to keep the original message when possible, which dictates the `try ... except` approach in the PR.

### DIFF
--- a/yamale/command_line.py
+++ b/yamale/command_line.py
@@ -120,10 +120,13 @@ def main():
     args = parser.parse_args()
     try:
         _router(args.path, args.schema, args.cpu_num, args.parser, not args.no_strict)
-        print('Validation success! 👍')
     except (SyntaxError, NameError, TypeError, ValueError) as e:
         print('Validation failed!\n%s' % str(e))
         exit(1)
+    try:
+        print('Validation success! 👍')
+    except UnicodeEncodeError:
+        print('Validation success!')
 
 
 if __name__ == '__main__':


mildebrandt:
Hi, thanks for your interest in Yamale!

Can you provide some recreation steps? I ran Yamale within an Ubuntu docker container and the "Validation success! 👍" message prints without error.

sirykd:
> Can you provide some recreation steps? I ran Yamale within an Ubuntu docker container and the "Validation success! 👍" message prints without error.

To clarify, the issue is image-dependent if we talk about Docker (e.g. [generating locale can be a hot-fix](https://medium.com/quiq-blog/handling-emoticons-or-other-unicode-characters-beware-of-docker-images-f0f11673dac4)).

Please check the following Dockerfile to reproduce it:

```
FROM ubuntu:18.04

RUN apt-get update && apt-get upgrade -y && apt-get clean
RUN apt-get install -y python3-pip
RUN pip3 install yamale
```

Build an image and run yamale:

```bash
docker build -t unicode-fail-image .
docker run unicode-fail-image yamale
```

Expected result:

```
Validation failed!
'ascii' codec can't encode character '\U0001f44d' in position 20: ordinal not in range(128)
```

mildebrandt:
Well that's unfortunate, I'm surprised it took 5 years for someone to have an issue with that. Thanks for the recreation steps.

mildebrandt (approved):

mildebrandt:
Your fix is in release 4.0.3, thanks for the contribution!

sirykd:
> Your fix is in release 4.0.3, thanks for the contribution!

Thanks!

PR #192, opened 2022-02-16
Support relative schema path

jtyr:
This PR is adding a check if the specified schema is actually a path to an existing file. Without this change, the schema cannot be find if the data path and the schema path combination doesn't produce an existing path to the schema. To demonstrate this, let's have this directory structure:

```
/dir1/data/myfile.yaml
/dir2/schemas/myfile.yaml
```

If we are in the `/dir2` and we run Yamale, the schema is not found:

```shell
cd /dir2
yamale -s ./schemas/myfile.yaml /dir1/data/myfile.yaml
```

### DIFF
--- a/yamale/command_line.py
+++ b/yamale/command_line.py
@@ -51,6 +51,8 @@ def _find_schema(data_path, schema_name):
     """ Checks if `schema_name` is a valid file, if not
     searches in `data_path` for it. """
 
+    if os.path.isfile(schema_name):
+        return schema_name
     directory = os.path.dirname(data_path)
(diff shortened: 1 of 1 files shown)


mildebrandt:
Thanks! This may cause a breaking change to some use cases....I'll research a bit.

jtyr:
As you look into this, please could you also look into the other PR #171 I have create a while ago?

cblakkan:
This was fixed in another PR. Thanks for the contribution, sorry it took so long :)

Issue #166, opened 2021-08-04
Feature Enhancement: Using data from the environment with validation


gms8994:
Imagine I have files in several folders, where each folder was a US state name (tennessee, virginia, massachusetts), and inside each of them contained a yaml file with the following contents:

```
state_name: "<state name>"
state_url: "https://<state name>.example.com"
```

I'd essentially like to be able to call yamale by setting an environment variable (`STATE=tennessee yamale state.yaml`) and have that value be the only thing allowed. Similarly, if I could do some form of string concatenation such that I could confirm that environment variable were provided as part of my value, that would be excellent. Any thoughts on this enhancement request? Possible example of the resulting schema.yaml below.

```
state_name: enum(env(STATE))
state_name: enum("https://" + env(STATE) + ".example.com")
```

mildebrandt:
I'm a little hesitant to add support for environment variables in the schema. We'd like to keep the schema as static as possible. Feel free to create your own validator, there's a small example here: https://github.com/23andMe/Yamale#custom-validators

And to support concatenation, you can override the `__add__` method on the validator class. Let me know if you need any help, and thanks for your interest in Yamale.

gms8994:
Is there an easy way to add a custom validator that I can then use directly from the commandline? It seems like no, but if that's the case, then it feels like the cli is limited only to "base" scenarios - is that the case?

mildebrandt:
That's correct, there is no current way to load custom validators from the command line. We would be open to having that feature, and it has been brought up before: https://github.com/23andMe/Yamale/issues/57

If you would like to implement this feature, we do accept pull requests. Please submit a small outline of how the user would use this feature before developing to ensure we're on the same page.

gms8994:
Good to know. I'll definitely consider a PR. As for outline... My initial thought would be something like:

Using the [literal style](https://yaml.org/spec/1.2/spec.html#style/block/literal) of YAML, allow "embedding" Python code directly in the schema file. This would give give the ability to keep the code and schema close together (though would prevent re-use of the code unless some kind of external include functionality were considered within the schema definition itself). You could then have something like:

```
custom_validator:
    type: validator
    code: |
        """
        Full contents of the CustomValidator(Validator) class goes here
        Where the class would be generated based on the name of the element (custom_validator) here
        And exec()'d to actually import the class
        Immediate checks could be done to determine if the eleemnt or the tag overloads any built-ins
        """
        
        tag = 'not_bool'

        def _is_valid(self, value):
[...]

PR #202, opened 2022-08-12
Command line argument for custom validators

jhe-iqbis:
First of all, thank you for this great tool. It is a crucial part of our build pipeline to maintain code quality and therefore incredibly useful to our project.

As Yamale already features a modular design, able to be extended by custom validators, my pull request aims to allow this using command line arguments.

There are no breaking changes in my pull request and I wrote a test that completes successfully for python 3.6, 3.8 and (with a newer pytest version) 3.10 on my system.

### DIFF
--- a/yamale/command_line.py
+++ b/yamale/command_line.py
@@ -105,6 +105,8 @@ def _validate_dir(root, schema_name, cpus, parser, strict):
 def _router(root, schema_name, cpus, parser, strict=True, include=None):
     root = os.path.abspath(root)
     if include is not None:
+        if not os.path.isfile(include):
+            raise ValueError("Python include file '{}' not found.".format(include))
         from importlib.machinery import SourceFileLoader
         SourceFileLoader(os.path.basename(include), include).load_module()
         yamale.validators.validators.update_default_validators()
@@ -126,7 +128,7 @@ def main():
     parser.add_argument('-p', '--parser', default='pyyaml',
                         help='YAML library to load files. Choices are "ruamel" or "pyyaml" (default).')
     parser.add_argument('-i', '--include', default=None,
-                        help='Path of Python library to load for custom validators.')
+                        help='File path of Python library to load for custom validators.')
     parser.add_argument('--no-strict', action='store_true',
                         help='Disable strict mode, unexpected elements in the data will be accepted.')
     args = parser.parse_args()


mildebrandt:
Thanks for your contribution! We usually like to have a discussion in an issue first to ensure not too much coding happens in the wrong direction. This seems pretty contained though, so not a problem. Just to keep in mind for the future. 

I'll be out for a week, so it'll take a while for me to get to this. Initially, on the surface, it does look good though.

mechie:
If this is accepted/merged, the new flag should be documented in the README, probably along with a section that links to the example custom validators and a shows how to use it.

mechie:
Just going to quote from https://github.com/23andMe/Yamale/issues/166#issuecomment-895585284 here, since that issue is the mirror of this one (custom validators inlined in schemas), and the concerns mostly apply:
> Some things to think about:
> - What happens when the user mistakenly adds two validators with the same tag? Is that an error or a warning?
> - What happens when the user attempts to override an existing validator?
> - Will this work in conjunction with loading a separate custom validator via code?
> - Keep in mind the other implementation in case we'd like to do that in the future as well. I'd hate to totally rewrite this to fit the other implementation.
> - Tests, tests, tests....someone else will be maintaining your code.

mildebrandt:
I appreciate the reminder of the previous concerns. I believe this is a little different since it's not embedding the validators in the schema itself, but it's in a separate python file that is pointed to within the command line options. This makes it far less dangerous as well, and shouldn't require an extra warning in the README. 

However, I've only looked at this for 5 minutes, so please let me know if I missed something. :)

mechie:
Ah yep, same, just did an initial read and the non-technical concerns popped in my head, will need to come back and look deeper. I consume yamale only as a library, so the concern is not personal.

I'm not sure what promises/guarantees the CLI makes (is there an implicit "only the schema flag is dangerous" due to the warning?). From a user POV it's "I give flag A file X, and flag B file Y" so warning about A (schema) but not B (includes) when both can lead to undefined behavior (schema via ast exploits, includes via it being loaded as a python module) seems odd.

mildebrandt:
The danger comes from offering Yamale as a service. If that service is accepting a schema/data pair and offering back a result, there are concerns that processing uncontrolled schema would leave the service open to malicious attacks. 

When limited to command line, the caller of the `yamale` executable has full control over the additional validators that are imported via the command line option.

I understand your concern, and I don't mind putting a line in the README saying to "ensure the imported validators come from trusted sources" or something similar.

mildebrandt:
I wanted to do a little check in here. Work is really busy at the moment. This hasn't been forgotten, but will take a bit for me to get to....my apologies.

PR #226, opened 2023-07-31
Add URL validator

### LINKED ISSUE #224: Add URL validator

vascoguita:
Closes #224


mildebrandt:
Hi! I'm no longer the maintainer, but a few things stood out: 

- In the past, we've worked to keep the dependencies down to just the yaml related packages. Adding the `requests` package would be unnecessary for the majority of yamale users. 
- The only protocols supported are http and ftp...it would be good to see all possible protocols supported.
- Your functional test isn't running. See `test_data` and `test_bad_subset2()` in `test_functional.py` for an example. 
- The regex doesn't accept some URLs....such as `http://xn--nw2a.xn--j6w193g/`

Good luck getting this in.

vascoguita:
Hello @mildebrandt, thank you for your feedback.
I'll try to sort out the points you've mentioned and hopefully we can have this merged soon :)

vascoguita:
Hello again @mildebrandt,
It's possible to check if a URL is reachable using just the Python Standard Library (without `requests`).
However it will be hard to add a URL validator that accepts all possible URLs without bloating Yamale with a big regular expression or without adding a dependency such as [validators](https://github.com/python-validators/validators).

I think, a more satisfactory solution is to use this URL validator as a custom validator, without merging it to Yamale. I'm therefore closing this PR.

PR #232, opened 2023-11-01
avoid appending undefined path

junaid-ali:
When the path is undefined, we get an error msg starting with colon:
```shell
`: my error msg`
```


junaid-ali:
@mildebrandt if you can have a look please

PR #244, opened 2024-03-20
Add ruff gh action

cblakkan:
* Add ruff pre-commit hook and github action
* Fixes ruff lint errors. 
* Ignoring  `import *` and unused import warnings until there's more time to clean them up.


joecackler (approved):
maybe worth mentioning in the dev section of the readme that we use ruff to format/lint and that devs can install pre-commit and then `make install-hooks` to have this done automatically

PR #246, opened 2024-04-22
add semantic versioning (semver.org) validator

thiagowfx:
https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string


thiagowfx:
If you deem this PR worthwhile, please accept the workflow so that we can verify whether the tests pass.

cblakkan:
Looks like you've got a typo in the test

```

      def test_semver():
          """
          - https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
          - https://regex101.com/r/Ly7O1x/3/
          """
  >       v = val.Semver()
  E       AttributeError: module 'yamale.validators' has no attribute 'Semver'. Did you mean: 'SemVer'?
  
```

thiagowfx:
Ooops! Fixed

README.md
```diff
@@ -400,6 +400,12 @@ Validates MAC addresses.
 Examples:
 - `mac()`: Allows any valid MAC address
 
+### SemVer (Semantic Versioning) - `SemVer()`
+Validates [Semantic Versioning](https://semver.org/) strings.
+
+Examples:
+- `SemVer()`: Allows any valid SemVer string
```
cblakkan, line 407: I think this should be `semver()` instead of `SemVer()` right?

thiagowfx: Looking at the rest of the values, for consistency, yes. Updated.

Although, intuitively, I'd have expected the UpperCamelCase version to be the correct one, to mirror the class definition. I suppose it uses the value of `tag` then.

PR #249, opened 2024-06-07
Add ruamel dependency

emmeowzing:
project is missing the `ruamel.yaml` dependency. I had to install it manually after a failed runtime exception.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,10 @@ setup(
     license="MIT",
     packages=find_packages(),
     include_package_data=True,
-    install_requires=["pyyaml"],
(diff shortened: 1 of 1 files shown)


cblakkan:
ruamel.yaml is currently intended to be an optional dependency. The library should only depend on it if you specify the ruamel parser explicitly not by default. It would be cool if we supported something like `pip install yamale[ruamel]` where a user could choose to have ruamel.yaml be a dependency and pyyaml not be a dependency but currently that's not supported.

cblakkan:
Closing with #259

PR #256, opened 2024-11-15
add URL validator implementation & tests

bieli:


requirements.txt
```diff
@@ -1 +1,2 @@
 tox==4.13.0
+validators==0.34.0
```
mildebrandt, line 2: Hi! I have a few concerns about using the validators package:

1. It's still pre 1.0, so things can break between minor versions. https://github.com/python-validators/validators/issues/285#issuecomment-1676615432
2. It doesn't support `http://localhost` by default. https://github.com/python-validators/validators/issues/392
3. It doesn't support query strings without values. https://github.com/python-validators/validators/issues/363
4. The supported schemes are hard coded. https://github.com/python-validators/validators/blob/c9585e91f8b409029b059df148da4dc52bd951e3/src/validators/url.py#L42

And that's all I could find in the last 10 minutes. I'm sure this could work for a subset of users....but it's not a generic solution for all valid URLs.

bieli: Sure. Thanks for this feedback. I'll find something more universal.

mildebrandt: Cool. FYI, currently Yamale has only one dependency....`pyyaml`. Anything you add will double the number of dependencies. I don't know how the current maintainers feel about that, but just something to keep in mind. 

A couple other things:

1. Don't update the version of Yamale in your PR. That's done during the release process.
2. Add any new dependencies in the setup.py file.

Good luck!

PR #263, opened 2025-02-06
feat: Pass field names to Validators and Constraints

nbaju1:
This PR adds the ability for Validators and Constraints to access the field name they're validating. This enables more descriptive error messages and warnings that can reference the specific field being validated.

## Changes
- Added `field_name` property to `Validator` and `Constraint` classes
- Pass field name through validation chain
- Updated tests to verify field name accessibility
- Updated documentation

## Use Case
This is particularly useful for constraints that need to reference the field name in their messages, such as deprecation warnings or field-specific validation errors.


README.md
```diff
@@ -446,6 +446,38 @@ Validates included structures. Must supply the name of a valid include.
 Examples:
 - `include('person')`
 
+### Field Names in Constraints
+Constraints can now access the field name they're validating, enabling more descriptive error messages. This is particularly useful for custom constraints that need to provide field-specific messages.
```
mildebrandt, line 450: Consider changing to "Constraints can access the field name they're validating, enabling descriptive error messages". Removes the words "now" and "more". 

Since this is a README, the text should indicate the feature has always been there....since this text will be there for the next 10 years. :)

yamale/validators/constraints.py
```diff
@@ -13,6 +13,8 @@ class Constraint(object):
     is_active = False
 
     def __init__(self, value_type, kwargs):
+        self.field_name = None # Added for custom field-aware messages
+        self.value_type = value_type  # Added for custom type-aware messages
```
mildebrandt, line 17: Consider creating a test for `value_type`.

mildebrandt:
When I ran the tests, I got errors similar to the following:
```
>           validator.field_name = key                                                                                               
E           AttributeError: 'dict' object has no attribute 'field_name'                                                                                          
yamale/schema/schema.py:78: AttributeError                  
```

Did you get these errors when running the tests?

Unfortunately, not all validators are of the `Validator` class. They can also be `dict`, `list`, and probably others that escape my mind. If you wrap the setting of `field_name` in something like `if isinstance(validator, val.Validator):`, that should get you what you want.

nbaju1:
@mildebrandt, thanks for the review. Updated PR with your suggestions.

README.md
```diff
@@ -446,6 +446,38 @@ Validates included structures. Must supply the name of a valid include.
 Examples:
 - `include('person')`
 
+### Field Names in Constraints
```
mildebrandt, line 449: I don't think there's a case where we can have a custom constraint without a custom validator....so I think moving this under the `Custom Validators` section may be better. Please move it down and indent the heading by one...so 4 # signs.

mildebrandt:
I think this looks good. But I have no power here anymore. :) 

Someone from the team will want to review and decide if it's a feature that they want to include in Yamale.

nbaju1:
@cblakkan, did you get a chance to look at this?

cblakkan:
> I like the "deprecated" flag idea, but it would be more informative if it could be a message, not just a boolean value.
> 
> ```
> password: str(deprecated=True)
> password: str(deprecated="The `password` field is deprecated. Please replace with SSO config")
> ```

I've been busy and haven't had a chance to dig into this yet but this was suggested to me by @joecackler and if done this way you would just need to write your own `Constraint` and there wouldn't need to be a change to Yamale.

> This enables more descriptive error messages and warnings that can reference the specific field being validated.

One of the things I was going to dig into was how helpful the change might be in improving the existing validation error messages. Though if we do pass field_name and value_type to the Constraint it seems like it should be passed via the `is_valid` function as arguments instead of mutating the object.

sambowry:
I think the field name check should be a constraint of all validators. I'm experimenting with it, because it help to validate some map items strictly.

For example:
```yaml
data: map( include('item') )
---
item: any( include( 'item_strict', name=str(equals='item_name') ),
           include( 'item_loose',  name=str(matches='^.*(?<!^item_name)$')) )
... etc ...
```
In the future, it could look like this:
```yaml
item: any( include( 'item_strict', name='item_name'      ),
           include( 'item_loose',  name=not('item_name') ) )
```

Issue #264, opened 2025-03-02
map( key=inlclude() ) validate all keys


sambowry:
I tried to use str=include() in map() to add meaningful names to the scheme, but it looks that the include() is not evaluated and it validates all keys.

test.yml
```
data:
  a: 1
  b: 2
  c: 3
```
schema.yml
```
# not correct, all keys are valid
data: map( int(), key=include('letter_a') )

# correct, only 'a' is valid
# data: map( int(), key=str( matches='^[a]$' ) )

---
letter_a: str( matches='^[a]$' )
```

sambowry:
Now it works for me.
```
class Schema(object):
  . . .
    def __init__(self, schema_dict, name="", validators=None, includes=None):
        . . .
        yamale.schema = self

class Include(Validator):
  . . .
    def _is_valid(self, value):
        return not yamale.schema.includes[self.include_name].validate( value, self.include_name, True ).errors

### test schema:
data:
  first:  map( int(), key=include('class_name') )
  second: map( int(), key=include('letter_b'  ) )
---
letter_a: str( matches='^[a]$' )
letter_b: str( matches='^[b]$' )
class_name: include('letter_a')

### test data:
data:
  first:
    a: 1
    b: 2
    c: 3
  second:
    a: 1
    b: 2
    c: 3

### result:
Validating test.yml...
Validation failed!
Error validating data 'test.yml' with schema 'test-schema.yml'
        data.first: Key error - 'b' is not a class_name.
        data.first: Key error - 'c' is not a class_name.
        data.second: Key error - 'a' is not a letter_b.
        data.second: Key error - 'c' is not a letter_b.
```

PR #266, opened 2025-03-08
bugfix/map( key=inlclude() ) validate all keys

sambowry:
in map( key=include() )  the include() is not evaluated and it validates all keys

https://github.com/23andMe/Yamale/issues/264


cblakkan:
Thanks for the submission, will take a look as soon as I get some time.

PR #273, opened 2025-04-25
[feature] all() validator

### LINKED ISSUE #272: [feature] all validator
similar to having any() validator would be nice to have all() of a list of validators

qequ:
Fixes #272

### DIFF
--- a/yamale/tests/fixtures/all.yaml
+++ b/yamale/tests/fixtures/all.yaml
@@ -1,0 +1,6 @@
+number_range: all(num(min=5), num(max=10))
+password: all(str(min=8), regex('^(?=.*[a-z])(?=.*[A-Z])(?=.*\d).*$'))
+email: all(str(), regex('^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$'))
+mixed_list: list(all(num(min=0), num(max=100)))
+nested:
+  value: all(int(min=10), int(max=20))
--- a/yamale/validators/validators.py
+++ b/yamale/validators/validators.py
@@ -157,6 +157,31 @@ class Any(Validator):
         return True
 
 
+class All(Validator):
+    """All validators must succeed validator"""
+
+    tag = "all"
+
+    def __init__(self, *args, **kwargs):
+        self.validators = [val for val in args if isinstance(val, Validator)]
+        super(All, self).__init__(*args, **kwargs)
+
+    def _is_valid(self, value):
+        return True
+
+    def is_valid(self, value):
+        # Override to validate against all validators
+        if not self._is_valid(value):
+            return False
+
+        # All validators must succeed
+        for validator in self.validators:
+            if not validator.is_valid(value):
+                return False
+
+        return True
+
+
 class Subset(Validator):
     """Subset of several types validator"""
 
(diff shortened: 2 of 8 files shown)


qequ:
> Thank you for the PR! I'm not convinced that `all()` is a necessary addition to the library, though. It looks like the example use cases reflected in your test data can each be achieved more simply with existing validators. Do you have other use cases in mind for which `all()` would be essential?

The motivation behind this PR actually came from a use case I had where I needed to compose several heterogeneous validators — and many custom ones — that couldn’t be easily combined using existing syntax.

While the test examples in the PR are simple, they were intended to verify core functionality. In practice, I ran into a case where I needed to validate a field with a str and several custom validators that made extra checks.

There was no straightforward way to express this combination without all(), especially when custom logic is involved. Wrapping this validation in all() made it readable and extensible.

I'm happy to update the test cases to reflect this more realistic scenario if you think that would help clarify the value of this addition.

yamale/tests/fixtures/all.yaml
```diff
@@ -0,0 +1,6 @@
+number_range: all(num(min=5), num(max=10))
+password: all(str(min=8), regex('^(?=.*[a-z])(?=.*[A-Z])(?=.*\d).*$'))
+email: all(str(), regex('^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$'))
+mixed_list: list(all(num(min=0), num(max=100)))
+nested:
+  value: all(int(min=10), int(max=20))
```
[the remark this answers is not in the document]
qequ: I agree the examples in the tests might give the impression that all() is redundant with existing validators. My goal with those cases was more to confirm that the implementation was working correctly and integrating cleanly with the existing system, rather than to showcase complex validation logic.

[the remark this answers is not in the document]
joecackler: Thanks. To be clear, I don't think it's necessary to "showcase complexity", but rather to demonstrate utility. I agree there's a use case here involving composition of custom validators -- thanks for adding that to the README.

yamale/validators/validators.py
```diff
@@ -157,6 +157,31 @@ def _is_valid(self, value):
...
+    tag = "all"
+
+    def __init__(self, *args, **kwargs):
+        self.validators = [val for val in args if isinstance(val, Validator)]
+        super(All, self).__init__(*args, **kwargs)
+
+    def _is_valid(self, value):
+        return True
+
+    def is_valid(self, value):
```
joecackler, line 172: It seems perhaps better to override the `validate` method (which is called by `is_valid`) instead, along the lines of

```py
def validate(self, value):
    errors = []
    for validator in self.validators:
        errors += validator.validate(value)
    return errors
```

As it stands, `all_instance.is_valid(value)` can return `False` while `all_instance.validate(value)` returns no errors, which seems inconsistent.

yamale/validators/validators.py
```diff
@@ -157,6 +157,48 @@ def _is_valid(self, value):
...
+            return errors
+
+        # Then validate all the constraints
+        for constraint in self._constraints_inst:
+            error = constraint.is_valid(value)
+            if error:
+                if isinstance(error, list):
+                    errors.extend(error)
+                else:
+                    errors.append(error)
```
joecackler, line 193: Perhaps better to delegate everything above to the base class, ie something like:

```py
errors = super().validate(value)

# Now validate against all child validators
for validator in self.validators:
    errors.extend(validator.validate(value))
return errors
```

sambowry: Yes, the collection of error messages seems strange, especially when we look the NotAny() validator which should print out success messages.

PR #286, opened 2026-05-12
[codex] Remove eval from schema parser

anansutiawan:
## Summary
- replace the final schema parser `eval(compile(...))` step with a recursive AST constructor
- keep the existing validated grammar for constants, validator-name references, and nested validator calls
- add parser regression coverage for nested validator-name references and unary numeric constants

## Rationale
PR #173 added AST validation as a best-effort hardening step and noted that removing `eval` entirely would be the stronger follow-up. This keeps the same schema syntax while avoiding runtime evaluation of schema expressions.

## Validation
- `pytest yamale` (169 passed)
- `ruff check yamale/syntax/parser.py yamale/syntax/tests/test_parser.py`
- `bandit yamale/syntax/parser.py -q`

### DIFF
--- a/yamale/syntax/parser.py
+++ b/yamale/syntax/parser.py
@@ -2,9 +2,6 @@ import ast
 
 from .. import validators as val
 
-safe_globals = ("True", "False", "None")
-safe_builtins = dict((f, __builtins__[f]) for f in safe_globals)
-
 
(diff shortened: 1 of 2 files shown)


anansutiawan:
Closing this draft and reopening from a cleaner branch and title.

